PR #1, opened 2024-02-29
fix: kubebuilder generate fix

mateusoliveira43:
Fix generate rollbacks not applied in https://github.com/migtools/oadp-non-admin/commit/58e1d161661be90f038cc866917d026a9162666c


api/v1alpha1/groupversion_info.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package v1alpha1 contains API Schema definitions for the oadp.openshift.io v1alpha1 API group
+// Package v1alpha1 contains API Schema definitions for the nac v1alpha1 API group
```
mpryc, line 17: Should this be `for the nac in the oadp.openshift.io` API group? `v1alpha1` is only version of the API

go.mod
```diff
@@ -1,21 +1,18 @@
-module github.com/openshift/oadp-non-admin
+module github.com/migtools/oadp-non-admin
 
 go 1.21
 
 require (
+	github.com/go-logr/logr v1.4.1
 	github.com/onsi/ginkgo/v2 v2.14.0
 	github.com/onsi/gomega v1.30.0
```
mpryc, line 8: I've added anything manaualy added post generate to second group, is that wrong?

api/v1alpha1/groupversion_info.go
mateusoliveira43: I do think so, `oadp.openshift.io` is the domain, `nac` is the group (and `v1alpha`is version). But this is generated string

go.mod
mateusoliveira43: without comments yes, we can not know why they are separate

api/v1alpha1/groupversion_info.go
shubham-pampattiwar: +1 to what @mpryc is saying, @mateusoliveira43 what do you mean by domain ? The group is `nac.oadp.openshift.io` , version is `v1apha1`.
Ref link: https://github.com/migtools/oadp-non-admin/blob/1d7eac38b853fd776cf4a7a045f760b2710ac53e/api/v1alpha1/groupversion_info.go#L29

mateusoliveira43: looking the PROJECT file in the root of the repo

the string may be wrong, but its generated using PROJECT.group and PROJECT.version

group from kubernetes is generated using PROJECT.group + PROJECT.domain

PR #3, opened 2024-02-29
feat: add editorconfig

mateusoliveira43:
## Description

Add [editorconfig](https://editorconfig.org/) to project. This will check all project's file format, by checking:
- if there is no trailing whitespace in them
- if there is a empty newline at the end of them
- if they use the correct indent style (tab or space)
- if they have correct indent size

## How to test

Run `make ec`


mpryc (approved):

.editorconfig
```diff
@@ -0,0 +1,20 @@
+# https://EditorConfig.org
+
+root = true
+
+[*]
+indent_size = 4
+indent_style = space
```
kaovilai, line 7: For .go
> Indentation
> We use tabs for indentation and gofmt emits them by default. Use spaces only if you must.

https://go.dev/doc/effective_go#formatting

kaovilai (changes requested):

.editorconfig
mateusoliveira43: did not understand the linked PR

but yes, go files and Makefile both accept tabs and spaces (go example of spaces: the headers copyright), so added them to list of  `indent_style=unset`. `go fmt` (run by golangci-lint) will take care of this for us in go files

kaovilai: > [The tab character is a very important part of the syntax of Makefiles.](https://stackoverflow.com/a/2131227/4590470)

So Makefile should not be indented using spaces. Yes unset sounds good.

Spaces can exists in go and makefile but not as an indent method for non-comments.

Makefile
```diff
@@ -204,8 +204,23 @@ endef
...
+	ec_tar=$(LOCALBIN)/$${ec_binary}.tar.gz ;\
+	curl -sSLo $${ec_tar} https://github.com/editorconfig-checker/editorconfig-checker/releases/download/$(EC_VERSION)/$${ec_binary}.tar.gz ;\
+	tar xzf $${ec_tar} ;\
+	rm -rf $${ec_tar} ;\
+	mv $(LOCALBIN)/$${ec_binary} $(EC) ;\
+	}
+
 .PHONY: ci
-ci: simulation-test lint docker-build hadolint check-generate check-manifests ## Run all checks run by the project continuous integration (CI) locally.
+ci: simulation-test lint docker-build hadolint check-generate check-manifests ec ## Run all checks run by the project continuous integration (CI) locally.
```
kaovilai, line 223: > go fmt (run by golangci-lint) will take care of this for us in go files

move lint to after ec
```suggestion
ci: simulation-test docker-build hadolint check-generate check-manifests ec lint ## Run all checks run by the project continuous integration (CI) locally.
```

kaovilai (changes requested):

kaovilai (approved):

Makefile
mateusoliveira43: `make ec` will only read

the vscode plugin and other stuff do the writing part, but did not find an executable, so we wont have `make ec-fix` command :(

PR #6, opened 2024-02-29
feat: add CI to project

### LINKED ISSUE #4: Add CI
Add a continuous integration solution to the project. It should run:
- tests
- linters
- formatters
- check if files are up to date (if `make manifests` and `make generate` were run)
- etc

mateusoliveira43:
Fixes #4 by adding GitHub actions to the project.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Go linters
+        run: make lint
+
+  container-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build container image
```
mateusoliveira43, line 44: there is no need to build the image, added just to check what you think. Should we keep?

mpryc: We should build the container and host it in quay I think.

I have added the following three github secrets:
```
QUAY_USERNAME
QUAY_PASSWORD
QUAY_IMAGE_NAMESPACE
```

To be uses same as:
   https://github.com/dora-metrics/pelorus/blob/master/.github/workflows/cd.yml#L79-L81

Which is using template:
   https://github.com/dora-metrics/pelorus/blob/master/.github/workflow_templates/build_push_exporters/action.yml

mpryc: @mateusoliveira43 I think we need to build it as the oadp manager will deploy that controller from the image itself, so it has to be build and hosted somewhere. I think the best now is to use latest during development.

mateusoliveira43:
@mpryc please merge, tested here https://github.com/mateusoliveira43/oadp-non-admin/pull/1 CI is working

mpryc (approved):

PR #10, opened 2024-03-01
fix: Enable more linters

mateusoliveira43:
Enable more strict checks with Go linters using golangci-lint.


.golangci.yml
```diff
@@ -1,40 +1,128 @@
+# Documentation reference https://github.com/golangci/golangci-lint/blob/v1.56.2/.golangci.reference.yml
```
mateusoliveira43, line 1: update version used

.golangci.yml
```diff
@@ -1,40 +1,128 @@
...
 run:
-  deadline: 5m
+  skip-dirs-use-default: false
+  modules-download-mode: readonly
   allow-parallel-runners: true
+  skip-dirs:
```
mateusoliveira43, line 6: will remove these skips prior to merge

.golangci.yml
```diff
@@ -1,40 +1,128 @@
...
+    - linters:
+        - revive
+      text: "^struct-tag: unknown option 'inline' in JSON tag$"
+    - linters:
+        - revive
+      text: "^add-constant: avoid magic numbers like '0', create a named constant for it$"
+    - linters:
+        - revive
+      text: "^add-constant: avoid magic numbers like '1', create a named constant for it$"
+    # - linters:
```
mateusoliveira43, line 120: will uncomment or remove prior to merge

.golangci.yml
mateusoliveira43: will let this one for another PR

we are not using e2e yet

internal/common/function/function.go
```diff
@@ -0,0 +1,224 @@
...
+	if nonAdminBackup.ObjectMeta.UID != types.UID(nabOriginUUID) {
+		return nil, fmt.Errorf("UID from annotation does not match UID of fetched NonAdminBackup object")
+	}
+
+	return nonAdminBackup, nil
+}
+
+// TODO import?
+// Similar to as pkg/common/common.go:AppendUniqueKeyTOfTMaps from github.com/openshift/oadp-operator
+func mergeUniqueKeyTOfTMaps[T comparable](userMap ...map[T]T) (map[T]T, error) {
```
kaovilai, line 203: This function needs more comment to quickly understand if latter maps overrides the previous one or vice versa.

internal/common/function/function.go
mateusoliveira43, line 1: here we could separate in more files, but did not focused on this in this PR

internal/common/function/function.go
mateusoliveira43: done

mpryc:
Thanks.

Issue #11, opened 2024-03-03
Remove VeleroBackup controller and replace it with watch


mpryc:
Having two controllers for one Kubernetes object leads to complexity, concurrency issues, and inefficient resource usage. Consolidating logic into a single controller and implementing watch functionality within it is a better approach.

Multiple controllers operating on the same object can introduce race conditions and concurrency issues. If both controllers try to update the object simultaneously, it can lead to inconsistent states or conflicts. We already have controller for VeleroBackup object.

This Issue is to remove VeleroBackup controller and include it's current Reconcile loop functionality within the NonAdminBackup controller.

mateusoliveira43:
commit that introduced it as reference https://github.com/migtools/oadp-non-admin/commit/5ff6f33855b8d8876dff48579fbb6f02cda23f7e

shubham-pampattiwar:
@mpryc Thanks for summing up our discussion. Adding some helpful links with regards to the issue fix:
Controller API watches: https://github.com/openshift/oadp-operator/blob/76f0049ebe2dd3cb4e5ed710553d76323a8f236b/controllers/dpa_controller.go#L149 
Predicate logic to filter out events based on labels: https://github.com/openshift/oadp-operator/blob/76f0049ebe2dd3cb4e5ed710553d76323a8f236b/controllers/predicate.go#L12 
Responsibilities of the Velero Backup controller that can be added as a task in Non-Admin Backup controller reconcile logic and thus removing Velero Backup controller altogether: https://github.com/migtools/oadp-non-admin/blob/065eed67867eecd4422b89fe6808172ca05c4316/internal/controller/velerobackup_controller.go#L53-L92

mpryc:
> commit that introduced it as reference [5ff6f33](https://github.com/migtools/oadp-non-admin/commit/5ff6f33855b8d8876dff48579fbb6f02cda23f7e)

@mateusoliveira43 thanks for this. The RBAC imo will need to stay as we will need to update the VeleroBackup from NonAdminBackup controller. So this can't be removed: https://github.com/migtools/oadp-non-admin/commit/5ff6f33855b8d8876dff48579fbb6f02cda23f7e#diff-6d4c4e29ec096149781536bb9118fde8242c1901df764e3eda3019ac8503d6c7

PR #12, opened 2024-03-05
Add Watch of Velero Backup and remove Backup controller

mpryc:
Change which removes Backup Controller and adds the Watch of the Backup object within NonAdminBackup Controller.

Added two main predicates with 'upper' logic within composite predicate to easily distinguish between predicate for NonAdminBackup and the Backup one.

The reconcile loop updates the Status and the Spec of the NonAdminBackup from the Velero Backup object. The Spec in the Velero Backup is updated by the Velero, so we use similar logic to update our NonAdminBackup one.

Fixes https://github.com/migtools/oadp-non-admin/issues/11


cmd/main.go
mateusoliveira43, line 1: does this need to reverted as well https://github.com/migtools/oadp-non-admin/commit/5ff6f33855b8d8876dff48579fbb6f02cda23f7e#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R54 ?

mpryc:
> Also, undo changes from PROJECT file [5ff6f33#diff-d1fb402f1269ba7396dca9cbd03b44ac055874669bce1d0f239fa2f54ea3584bR20](https://github.com/migtools/oadp-non-admin/commit/5ff6f33855b8d8876dff48579fbb6f02cda23f7e#diff-d1fb402f1269ba7396dca9cbd03b44ac055874669bce1d0f239fa2f54ea3584bR20)

We can't because we will get error:

```
ERROR	controller-runtime.source.EventHandler	kind must be registered to the Scheme	{"error": "no kind is registered for the type v1.Backup in scheme \"pkg/runtime/scheme.go:100\""
```

internal/controller/velerobackup_handler.go
mateusoliveira43, line 1: why this file? what is the difference between this and predicate?

mpryc: Yes. Handler handles events which are let by predicates. They are together but separate:

```
EVENT ---> PREDICATE (true) ----> HANDLER ----> DO something

EVENT ----> PREDICATE (false) ---> Do nothing
```

internal/controller/velerobackup_handler.go
```diff
@@ -0,0 +1,70 @@
...
+func (h *VeleroBackupHandler) Create(ctx context.Context, evt event.CreateEvent, q workqueue.RateLimitingInterface) {
+	nameSpace := evt.Object.GetNamespace()
+	name := evt.Object.GetName()
+	log := getVeleroBackupHandlerLogger(ctx, name, nameSpace)
+	log.V(1).Info("Received Create VeleroBackupHandler")
+}
+
+func (h *VeleroBackupHandler) Update(ctx context.Context, evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	nameSpace := evt.ObjectNew.GetNamespace()
+	name := evt.ObjectOld.GetName()
```
mateusoliveira43, line 35: why old here?

internal/controller/nonadminbackup_predicate.go
```diff
@@ -0,0 +1,67 @@
...
+		nameSpace := evt.Object.GetNamespace()
+		name := evt.Object.GetName()
+		log := getNonAdminBackupPredicateLogger(ctx, name, nameSpace)
+		log.V(1).Info("Received Delete NonAdminBackupPredicate")
+		return true
+	}
+	return false
+}
+
+func (predicate NonAdminBackupPredicate) Generic(ctx context.Context, evt event.GenericEvent) bool {
```
mateusoliveira43, line 58: generic would be the default of switch case?

internal/controller/nonadminbackup_controller.go
```diff
@@ -142,5 +148,13 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 func (r *NonAdminBackupReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&nacv1alpha1.NonAdminBackup{}).
+		Watches(&velerov1api.Backup{}, &VeleroBackupHandler{}).
```
mateusoliveira43, line 151: how we know the difference between update and patch?

internal/controller/velerobackup_predicate.go
```diff
@@ -0,0 +1,69 @@
...
+
+	return nameSpace == veleroBackupPredicate.OadpVeleroNamespace
+}
+
+func (veleroBackupPredicate VeleroBackupPredicate) Delete(ctx context.Context, evt event.DeleteEvent) bool {
+	// NonAdminBackup should not care about VeleroBackup delete events
+	return false
+}
+
+func (veleroBackupPredicate VeleroBackupPredicate) Generic(ctx context.Context, evt event.GenericEvent) bool {
```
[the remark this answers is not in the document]
mpryc: We need as we are implementing interface.

internal/controller/nonadminbackup_predicate.go
mpryc: Here you have more about those 4 predicates, it's interface which we need to implement:
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.2/pkg/predicate/predicate.go#L32

internal/controller/composite_predicate.go
```diff
@@ -0,0 +1,51 @@
...
+	return p.VeleroBackupPredicate.Delete(p.Context, evt)
+}
+
+func (p CompositePredicate) Generic(evt event.GenericEvent) bool {
+	// If NonAdminBackupPredicate returns true, ignore VeleroBackupPredicate
+	if p.NonAdminBackupPredicate.Generic(p.Context, evt) {
+		return true
+	}
+	// Otherwise, apply VeleroBackupPredicate
+	return p.VeleroBackupPredicate.Generic(p.Context, evt)
```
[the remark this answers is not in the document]
mpryc: We could but then it's harder to put some clear comments.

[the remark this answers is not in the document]
mpryc: Plus there may be other conditions for predicate for each of the objects, so having each block for each "type" makes sense imo.

internal/controller/nonadminbackup_controller.go
kaovilai: For events, everything is called update including patch :)

internal/controller/velerobackup_predicate.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (veleroBackupPredicate VeleroBackupPredicate) Create(ctx context.Context, evt event.CreateEvent) bool {
+	nameSpace := evt.Object.GetNamespace()
+	name := evt.Object.GetName()
+	log := getBackupPredicateLogger(ctx, name, nameSpace)
+	log.V(1).Info("Received Create VeleroBackupPredicate")
+
+	if nameSpace != veleroBackupPredicate.OadpVeleroNamespace {
+		return false
+	}
```
kaovilai, line 40:
```suggestion
	nameSpace := evt.Object.GetNamespace()
	if nameSpace != veleroBackupPredicate.OadpVeleroNamespace {
		return false
	}

	name := evt.Object.GetName()
	log := getBackupPredicateLogger(ctx, name, nameSpace)
	log.V(1).Info("Received Create VeleroBackupPredicate")
```

for performance, let's not call getName, getLogger, log if the namespace is different.

kaovilai (changes requested):
predicates should be very performant cause there can be many hundreds of these at any given time that need filtering out.

internal/controller/velerobackup_handler.go
mateusoliveira43: controller runtime calls it internally event handler https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2/pkg/handler#EventHandler and predicate https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2/pkg/predicate#Predicate

mpryc:
@mateusoliveira43 could we merge this now?

mateusoliveira43 (approved):
lets merge to unblock PRs, we can analyze code more in depth next week

PR #16, opened 2024-03-12
fix: Add contribution documenation to project

mateusoliveira43:
Add contribution documentation to the project.

Also:
- created `deploy-dev` command
- separated documentation in README to `docs` folder
- fixed minor problems in NAC code


config/default/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: oadp-nac-system
+namespace: oadp-nac-system # TODO change
```
mateusoliveira43, line 2: what should be our default namespace? I think I heard us think about `openshift-adp`?

mateusoliveira43: also, want to find a easy way to change this from makefile

config/rbac/kustomization.yaml
```diff
@@ -9,7 +9,6 @@ resources:
 - role_binding.yaml
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
-- nonadminbackup_editor_role.yaml
```
mateusoliveira43, line 12: I think this is not necessary, it is a subset of `config/rbac/role.yaml`. Am I right to remove this?

docs/architecture.md
```diff
@@ -0,0 +1,23 @@
...
+    --domain=oadp.openshift.io
+kubebuilder create api \
+    --plugins go.kubebuilder.io/v4 \
+    --group nac \
+    --version v1alpha1 \
+    --kind NonAdminBackup \
+    --resource --controller
+make manifests
+```
+> **NOTE:** The information about plugin and project version, as well as project name, repo and domain, is stored in [PROJECT](PROJECT) file
```
mateusoliveira43, line 21: wrong link, fix

docs/non_admin_user.md
mateusoliveira43, line 1: want to test this in an AWS cluster prior to merging

if anyone (not Michal, who already have it setup) wants to test as well, would be good!

README.md
```diff
@@ -152,156 +47,43 @@ deployment.apps/oadp-nac-controller-manager   1/1     1            1           3
...
 
-To check if project code was generated, run
-```sh
-make check-generate
-make check-manifests
-```
+| NAC version | OADP version |
+|-------------|--------------|
+| master      | master       |
+| v0.1.0      | v1.4.0       |
```
mateusoliveira43, line 85: maybe just add this after first release?

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+make deploy-dev
+```
+
+> **NOTE:** You can use a different development image by running `export DEV_IMG=quay.io/user/image-name:tag` prior to `deploy-dev` command.
+
+To create a non admin user to test NAC, check [non admin user documentation](non_admin_user.md).
+
+To uninstall the previously installed OADP NAC in your cluster, run
+```sh
+make deploy-dev
```
mateusoliveira43, line 33: wrong, undeploy, fix it

config/default/kustomization.yaml
mpryc: I think for the beginning it should *not* be openshift-adp, so we don't make mistakes where we bind this controller with the OADP one. Having at first separate namespace will ensure those two are separate and can be then installed together in one.

README.md
mpryc: We could sync those two once we have it ? 

E.g. 

OADP: v.1.4.y
NAC: v.1.4.x

so the x and y may be different but the main parts will need to be in sync?

mateusoliveira43: all other plugins (this is not exactly a plugin...) have different versions, do not know if syncing versions would be good

config/default/kustomization.yaml
mateusoliveira43: will add this responsibility to the oadp integration task

docs/architecture.md
```diff
@@ -0,0 +1,23 @@
+# Architecture
+
```
mateusoliveira43, line 2: TODO add doc about how NAC integrates with OADP

.github/workflows/oadp-compatibility-check.yml
```diff
@@ -0,0 +1,46 @@
+name: OADP version compatibility check
+
+on:
+  push:
+    branches:
+      - master
+      # add release branches
+  schedule:
+    - cron: '0 0 * * 1' # Every Monday at 00:00 UTC
+  # TODO remove, just for test
```
mateusoliveira43, line 10: TODO

kaovilai: After PR merge?

mateusoliveira43: before, just to test if the job works, but first need the oadp repo PR merged

docs/architecture.md
mateusoliveira43: add also requirements to command work (ex.: need to already have `RELATED_IMAGE_NON_ADMIN_CONTROLLER` in manager.yaml)

config/default/manager_oadp_patch.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
```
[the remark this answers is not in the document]
mateusoliveira43: TODO compare deployment from `make build-installer` and created by OADP code. They should be equal

config/manager/manager.yaml
```diff
@@ -2,39 +2,42 @@ apiVersion: v1
...
     app.kubernetes.io/name: deployment
-    app.kubernetes.io/instance: controller-manager
+    app.kubernetes.io/instance: non-admin-controller-manager
     app.kubernetes.io/component: manager
-    app.kubernetes.io/created-by: oadp-nac
-    app.kubernetes.io/part-of: oadp-nac
+    app.kubernetes.io/created-by: oadp-operator
+    app.kubernetes.io/part-of: oadp-operator
     app.kubernetes.io/managed-by: kustomize
+    component: non-admin-controller
```
mateusoliveira43, line 27: TODO remove, use control-plane label in OADP code

Makefile
```diff
@@ -1,6 +1,6 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= quay.io/konveyor/non-admin-controller:latest
```
weshayutin, line 3: quay.io/konveyor/oadp-non-admin:latest ?

mateusoliveira43: correct, will change

README.md
```diff
@@ -1,307 +1,51 @@
...
-
-To upgrade kubebuilder version, create kubebuilder structure using the current kubebuilder version and the upgrade version, using the same commands presented earlier, in two different folders. Then generate a `diff` file from the two folders and apply changes to project code.
-
-## Quality
-
-The quality checks of the project are reproduced by the continuos integration (CI) pipeline of the project. CI configuration in [`.github/workflows/ci.yml`](.github/workflows/ci.yml) file.
+For a better understanding of the project, check our [architecture documentation](docs/architecture.md).
 
-To run all checks locally, run `make ci`.
+## OADP version compatibility
```
shubham-pampattiwar, line 41: IMHO we do not need this, right ? We would not be doing separate NAC releases that are decoupled from OADP Operator release.

mateusoliveira43: we will probably not do releases

but this compatibility information I do believe is important to know where the code comes from and to test 

For example, when 1.4 release branch is created, this needs to be changed there https://github.com/migtools/oadp-non-admin/pull/16/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R260

Makefile
```diff
@@ -241,3 +245,40 @@ check-manifests: manifests ## Check if 'make manifests' was run.
...
+
+# A valid oadp-operator git repo fork (for example https://github.com/openshift/oadp-operator)
+OADP_FORK ?= openshift
+# A valid branch or tag from oadp-operator git repo
+OADP_VERSION ?= master
+# namespace to deploy development OADP operator
+OADP_NAMESPACE ?= openshift-adp
+
+.PHONY: deploy-dev
+deploy-dev: TEMP:=$(shell mktemp -d)
```
shubham-pampattiwar, line 265: Question: Do you need these `deploy-dev` and `undeploy-dev` commands ? 
you would not be using this controller independently, right ? Also, if you wan to run controller locally, shouldn't `make run` suffice ?

mateusoliveira43: As specified in command description "Build and push development controller image from current branch and deploy development OADP operator using that image" this does not just create controller deployment in cluster, but an OADP operator with an non-admin image/CRDs with all changes from the current branch of `oadp-non-admin` the command is run from

example: when we create restore controller, we can use this command to test if it works with OADP, without an OADP PR before hand

I would suggest `make build-installer` instead of `make run` for just non-admin controller tests, as documented in architecture https://github.com/migtools/oadp-non-admin/pull/16/files#diff-140eef3ba41bdcf401d507408084181f2c0ac627532b61e0f7906ea7cc926782R7

shubham-pampattiwar: For any integration specific change between OADP Operator and NAC we will definitely need a PR at both the repos and that too at the same time. That means we will be testing OADP Operator and NAC at the same time in these scenarios and OADP Operator already has a `make deploy-olm` command that will install current branch changes (Operator, CRD etc) on your cluster. (You would be using `unsupportedOverrides` for NAC controller image) . Inter-repository changes are not new in our OADP feature development, plugin changes are also of the similar category sometimes. This is the perspective that I am looking from and I believe `make deploy-olm` from OADP Operator should be sufficient. Let me know your thoughts. Do you agree and share similar sentiment ?

README.md
```diff
@@ -1,307 +1,51 @@
...
-Following are the steps to build the installer and distribute this project to users.
+## Description
 
-1. Build the installer for the image built and published in the registry:
+This open source controller adds the non admin feature to [OADP operator](https://github.com/openshift/oadp-operator). With it, cluster admins can configure which namespaces non admin users can backup/restore.
 
-```sh
-make build-installer IMG=<some-registry>/oadp-nac:tag
-```
+## Getting Started
```
mateusoliveira43, line 16: Add a sample application install/usage to test backup/restore

Use this app https://github.com/migtools/oadp-non-admin/pull/34

Makefile
mateusoliveira43: for me, this command is the most important thing of this PR, allowing us to easy and early test NAC

But I do agree about `make deploy-olm`, we will use it in this command!

Basically this command just do 3 things
- clone oadp (we could point to a path in developer computer, but I think this way is easier to test, you can point to different fork/branch with env vars)
- run new make command `update-non-admin-manifests` (which updates the crd/rbac/samples/image of NAC of that cloned oadp) (because of this, no need for `unsupportedOverrides`)
- run `make deploy-olm` from the cloned oadp

so, if your NAC PR do change something that `update-non-admin-manifests` does not take care of (for example, you change something in NAC deployment https://github.com/openshift/oadp-operator/pull/1370/files#diff-ece703e5852171fe350087ab8bd7d5e14c7b577a83664fadc7506cafd54117e2 ) then, a simultaneously PR in oadp can/will be needed. 

Otherwise, if you just updated some NAC CRDs fields, for example, you will not have to have 2 simultaneously PRs, only one here and this new command to test NAC with OADP.

Of course, after that PR merges in NAC, a PR will be needed in OADP, which will be enforced by https://github.com/migtools/oadp-non-admin/pull/16/files#diff-c71fe913c680bf7af72b808a7695833cdbca6a10950984c003bdfc52f69555c5 and you just need to run `make update-non-admin-manifests` from compatible oadp branch to create that PR

so, in other words, in my point of view, this command removes necessity of simultaneously and blocking PRs in NAC and OADP repo and makes a little more easy/flexible to test NAC, that's why I feel it is so important

does it make sense to have the command with this explanation?

Makefile
mateusoliveira43: done

.github/workflows/oadp-compatibility-check.yml
```diff
@@ -0,0 +1,42 @@
+name: OADP version compatibility check
+
+on:
+  push:
+    branches:
+      - master
+      # add release branches
```
mpryc, line 7: We can't do that prior to merge? e.g. PR ?

Makefile
```diff
@@ -1,6 +1,6 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= quay.io/konveyor/oadp-non-admin:latest
```
mpryc, line 3: We currently have:

 https://quay.io/repository/migtools/oadp-nac
 
 Should we use the above or create new one within migtools and not konveyor?

Makefile
```diff
@@ -241,3 +260,40 @@ check-manifests: manifests ## Check if 'make manifests' was run.
...
+deploy-dev: NAC_PATH:=$(shell pwd)
+deploy-dev: AUX_RANDOM:=$(shell echo $$RANDOM)
+deploy-dev: DEV_IMG?=ttl.sh/oadp-nac-controller-$(shell git rev-parse --short HEAD)-$(AUX_RANDOM):1h
+deploy-dev: ## Build and push development controller image from current branch and deploy development OADP operator using that image
+	IMG=$(DEV_IMG) make docker-build docker-push
+	git clone --depth=1 git@github.com:$(OADP_FORK)/oadp-operator.git -b $(OADP_VERSION)  $(TEMP)/oadp-operator
+	cd $(TEMP)/oadp-operator && \
+	NON_ADMIN_CONTROLLER_PATH=$(NAC_PATH) NON_ADMIN_CONTROLLER_IMG=$(DEV_IMG) OADP_TEST_NAMESPACE=$(OADP_NAMESPACE) \
+		make update-non-admin-manifests deploy-olm && cd -
+	chmod -R 777 $(TEMP) && rm -rf $(TEMP)
```
mpryc, line 290: Can we have clean-temp target? It's much safer to check first then chmod and rm:

```Makefile
TEMP_CLEANUP := $(TEMP)/oadp-operator

clean-temp:
	@echo "Cleaning up temporary files..."
	if [ -d "$(TEMP_CLEANUP)" ]; then \
		chmod -R u+w "$(TEMP_CLEANUP)" && rm -rf "$(TEMP_CLEANUP)"; \
			echo "Temporary files cleaned up."; \
	else echo "Temporary directory not found. No cleanup needed."; fi
```

Makefile
mateusoliveira43: we configured the above one in PROW https://github.com/openshift/release/blob/master/core-services/image-mirroring/konveyor/mapping_konveyor_latest#L41

Makefile
mateusoliveira43: Have a PR for this purpose https://github.com/openshift/oadp-operator/pull/1382

.github/workflows/oadp-compatibility-check.yml
mateusoliveira43: did not understand

mateusoliveira43: yes, that is intentional

example: we add restore CRD here, we also need to add in OADP

if this is run during PR, we need simultaneously PRs in oadp and here

this way, we do thing here, and then, only after PR here merges, we have to create PR in oadp

mpryc: Shouldn't this be like:
 - we know there is problem with sync, so we override merge?
 
 Like this we may miss post merge failure and then we need to revert back?

mateusoliveira43: If is during PR, we need to open simultaneously PR in oadp, and merge it prior to the one here, meaing oadp will have new NAC stuff prior to NAC. I think is wrong

I preffer to be like this, after merging here, you need to update OADP. more clear path and makes more sense, in my opinion

similar discussion https://github.com/migtools/oadp-non-admin/pull/16#discussion_r1556352449

docs/architecture.md
```diff
@@ -0,0 +1,50 @@
...
+## OADP integration
+
+Normally, to ship a controller to users, the project would present the file created by `make build-installer` command (which include various Kubernetes objects, like Namespace, ServiceAccount, Deployment, etc), to user to install the controller. But since NAC needs OADP operator to properly work, those Kubernetes objects are shipped within OADP operator (and also Kubernetes objects in `config/samples/` folder). Because of this restriction, generated Kubernetes objects names and labels in `config/` folder, may need to be updated to match OADP operator standards (and avoid duplications, by adding `non-admin-` prefix to Kubernetes object names).
```
mateusoliveira43, line 5: talk about label `oadp-nac` that needs to be changed to `oadp-operator` as example

Makefile
mateusoliveira43: remove this part of the code, resolving

README.md
mateusoliveira43: to avoid confusion, matrix was removed

README.md
mateusoliveira43: removed

Makefile
mateusoliveira43: removed command and added how to run from OADP

hack/samples/apps/mysql.yaml
```diff
@@ -0,0 +1,208 @@
...
+      namespace: ${NAMESPACE}
+      labels:
+        app: mysql
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 1Gi
+  # - kind: SecurityContextConstraints
```
mateusoliveira43, line 27: can this be deleted @weshayutin ? A non admin user can not create that

example error
```
Error from server (Forbidden): error when creating "STDIN": securitycontextconstraints.security.openshift.io is forbidden: User "batman" cannot create resource "securitycontextconstraints" in API group "security.openshift.io" at the cluster scope
```

README.md
```diff
@@ -1,307 +1,62 @@
...
 
 ## Contributing
-// TODO(user): Add detailed information on how you would like others to contribute to this project
 
-**NOTE:** Run `make help` for more information on all potential `make` targets
-
-More information can be found via the [Kubebuilder Documentation](https://book.kubebuilder.io/introduction.html)
+Please check our [contributing documentation](docs/CONTRIBUTING.md) to propose changes to the repository.
 
 ## Architecture
```
mateusoliveira43, line 56: Also tell about `docs/design/` folder here

mateusoliveira43: if https://github.com/migtools/oadp-non-admin/pull/46 get merged first; otherwise do there

README.md
```diff
@@ -1,307 +1,62 @@
...
-```sh
-kubectl apply -k config/samples/
-```
-
->**NOTE**: Ensure that the samples has default values to test it out.
+Non Admin Controller
 
-### To Uninstall
-**Delete the instances (CRs) from the cluster:**
+[![Continuos Integration](https://github.com/migtools/oadp-non-admin/actions/workflows/ci.yml/badge.svg)](https://github.com/migtools/oadp-non-admin/actions/workflows/ci.yml)
```
mrnold, line 5: Minor spelling edit, should be "Continuous". There are a few more in this pull request if you feel like fixing them.

mateusoliveira43: will change, thanks!

mateusoliveira43: done

hack/samples/apps/mysql.yaml
weshayutin: yes, it can be deleted

mateusoliveira43: done

weshayutin:
Looks good enough for where we currently stand w/ OADP non-admin.   Will update as necessary.
Thank you @mateusoliveira43

PR #34, opened 2024-04-05
add basic sample app that works w/ project admin

weshayutin:
* this app requires the non-admin user to be a namespace admin.
* namespace must be created by cluster-admin
* non-admin roles and role-bindings applied to namespace

### DIFF
--- a/test/sample-applications/default_sc.yaml
+++ b/test/sample-applications/default_sc.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: List
+items:
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
+      name: mysql
+      namespace: mysql-persistent
+      labels:
+        app: mysql
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 1Gi
--- a/test/sample-applications/mysql-persistent-csi.yaml
+++ b/test/sample-applications/mysql-persistent-csi.yaml
@@ -1,0 +1,182 @@
+apiVersion: v1
+kind: List
+items:
+  - kind: Namespace
+    apiVersion: v1
+    metadata:
+      name: mysql-persistent
+      labels:
+        app: mysql
+  - apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: mysql-persistent-sa
+      namespace: mysql-persistent
+      labels:
+        component: mysql-persistent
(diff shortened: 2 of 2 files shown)


test/sample-applications/default_sc.yaml
mrnold, line 1: Can this be renamed to something that more obviously hints about the contents? I'm not sure what "SC" would stand for in this context.

test/sample-applications/mysql-persistent-csi.yaml
```diff
@@ -0,0 +1,182 @@
...
+  - kind: SecurityContextConstraints
+    apiVersion: security.openshift.io/v1
+    metadata:
+      name: mysql-persistent-scc
+    allowPrivilegeEscalation: false
+    allowPrivilegedContainer: false
+    volumes:
+    - '*'
+    users:
+    - system: nacuser
```
mateusoliveira43, line 26: if created user has different name, it will work? (should this have space between : ?)

test/sample-applications/default_sc.yaml
mateusoliveira43: why 2 separated files?

mpryc:
Should there be an example of how to create non admin user (on top of below there is need to grant nacuser with ns admin)?

E.G.:

```shell
# Create test NS

$ oc create ns mysql-persistent
namespace/mysql-persistent created


# Create identity file with user nacuser that has password SamplePassword

$ htpasswd -c -B -b ./users_file.htpasswd nacuser SamplePassword
Adding password for user nacuser

# Create secret with the above htpasswd file in OpenShift

$ oc create secret generic htpass-secret --from-file=htpasswd=./users_file.htpasswd -n openshift-config
secret/htpass-secret created

# Create OAuth file and apply it OpenShift


$ cat > oauth-nacuser.yaml <<EOF
apiVersion: config.openshift.io/v1
kind: OAuth
metadata:
  name: cluster
spec:
  identityProviders:
  - name: oadp_nac_test_provider 
    mappingMethod: claim 
    type: HTPasswd
    htpasswd:
      fileData:
        name: htpass-secret 
EOF

$ oc apply -f oauth-nacuser.yaml

```

mateusoliveira43:
@weshayutin can we close this one and discuss in https://github.com/migtools/oadp-non-admin/pull/16 ? In last commit added mysql template

weshayutin:
feel free to close :)

shubham-pampattiwar:
Already merged

Issue #44, opened 2024-04-15
Design template


mateusoliveira43:
Create design template for files that will be stored in `docs/design/` folder.

PR #46, opened 2024-04-16
feat: Add design template

mateusoliveira43:
## Why the changes were made

Standardize repository designs.

Fixes https://github.com/migtools/oadp-non-admin/issues/44

## How the changes were made

Copied template from OADP operator repository https://github.com/openshift/oadp-operator/blob/master/docs/design/_template.md

Read template and see if it is clear about what developers must write.

### DIFF
--- a/docs/design/template.md
+++ b/docs/design/template.md
@@ -1,0 +1,53 @@
+# Design proposal template `<replace with your proposal's title>`
+
+_Note_: The preferred style for design documents is one sentence per line.
+*Do not wrap lines*.
+This aids in review of the document as changes to a line are not obscured by the reflowing those changes caused and has a side effect of avoiding debate about one or two space after a period.
+
+_Note_: The name of the file should follow the name pattern `<short meaningful words joined by '-'>_design.md`, e.g:
+`listener-design.md`.
+
+## Abstract
+One to two sentences that describes the goal of this proposal and the problem being solved by the proposed change.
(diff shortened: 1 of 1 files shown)


kaovilai (approved):

docs/design/template.md
```diff
@@ -0,0 +1,53 @@
+# Design proposal template `<replace with your proposal's title>`
+
+_Note_: The preferred style for design documents is one sentence per line.
+*Do not wrap lines*.
+This aids in review of the document as changes to a line are not obscured by the reflowing those changes caused and has a side effect of avoiding debate about one or two space after a period.
+
+_Note_: The name of the file should follow the name pattern `<short meaningful words joined by '-'>_design.md`, e.g:
+`listener-design.md`.
```
mrnold, line 8: Should "listener-design.md" be "listener_design.md"? Or should the rule be "<short meaningful words joined by '-'>-design.md"?

mateusoliveira43: on OADP we seem to not follow this rule (have with dash, underscore or no suffix)

since we are in `design` folder, I would say we do not need suffix

but I do not have any preferences

mrnold: I don't have a preference either, I just want the example and the rule to match. Taking out the whole note sounds even better, if no one is paying attention to it anyway.

mateusoliveira43: done, removed the sentence

mpryc (approved):

PR #42, opened 2024-04-09
feat: Add Non Admin Restore controller

mateusoliveira43:
## Why the changes were made

Non admin users want to restore NonAdminBackups. This PR adds NonAdminRestore CRD along with its controller.

Related to https://github.com/migtools/oadp-non-admin/issues/35

~Blocked by https://github.com/migtools/oadp-non-admin/pull/16~
~Blocked by https://github.com/migtools/oadp-non-admin/pull/54~
Blocked by https://github.com/migtools/oadp-non-admin/pull/67
Blocked by https://github.com/openshift/oadp-operator/pull/1388

## How the changes were made

Get kubebuilder version `3.14.0` from https://github.com/kubernetes-sigs/kubebuilder/releases and run command documented in Architecture https://github.com/migtools/oadp-non-admin/commit/62d521341c6c10310eecc96be9291327e66ec9b2

Run `make manifests`

Fix code quality issues

## How to test the changes made

TODO

## TODO

This PR still misses
- write restore logic
- [check PR size, can be separate] Adding example of configuring DPA with restricted NonAdminRestore fields https://github.com/migtools/oadp-non-admin/blob/94c4be9bf93da2d791281d2aaca54655d2821316/README.md?plain=1#L29
- https://github.com/migtools/oadp-non-admin/blob/94c4be9bf93da2d791281d2aaca54655d2821316/README.md?plain=1#L50
  - show user how to see NonAdminRestore status
  - show user how to see NonAdminRestore queue 
- unit/integration tests for new controller. Some ideas
  - [ERROR] try to create NonAdminRestore from Velero Backup
  - [ERROR] try to create NonAdminRestore from non existing NonAdminBackup
  - [ERROR] try to create NonAdminRestore NonAdminBackup with restricted fields
  - [ERROR] try to create NonAdminRestore in non existing namespace
  - [SUCCESS] create NonAdminRestore from NonAdminBackup owned by user

PR #54, opened 2024-04-29
Reconcile loop rework

mpryc:
NonAdminBackup reconcile loop major rework
    
Rework of the reconcile loop to include batch reconcile.
    
Move assignment of the Log or Context outside of the reconcile loop.

# How to test

Note it's best to test this PR together with https://github.com/migtools/oadp-non-admin/pull/56 which includes namespace guard (case d) in this instructions.

1. Setup non-admin user as per instructions: https://github.com/migtools/oadp-non-admin/blob/master/docs/non_admin_user.md
2. Create multiple NonAdminBackup objects in the namespace for which the non-admin user have namespace admin rights. After each creation see if at the end of reconcile the NonAdminBackup is correctly updated with its status and if in some cases VeleroBackup object was created in the openshift-adp namespace (as an admin).

All of the below are in the non-admin namespace `nacproject`:

a) Case where there is no backupSpec:

```
kind: NonAdminBackup
metadata:
  name: example-nospec
  namespace: nacproject
spec: {}
```

Result should be NAB in backingOff phase with `BackupSpec is not defined` condition message:

```
kind: NonAdminBackup
metadata:
  name: example-nospec
  namespace: nacproject
spec: {}
status:
  conditions:
    - lastTransitionTime: '2024-05-02T08:24:50Z'
      message: BackupSpec is not defined
      reason: InvalidBackupSpec
      status: 'False'
      type: Accepted
  phase: BackingOff

```

b) Case where there is backupSpec, but empty, should create Velero Backup with nacproject namespace in the includedNamespaces (below is not full spec, removed `managedFields` from it):

```
kind: NonAdminBackup
metadata:
  name: empty-backupspec
  namespace: nacproject
spec: 
  backupSpec: {}
```

Result should be created Velero Backup with `includedNamespaces` same as the origin NAB namespace and updated NonAdminBackup (below is not full spec, removed `managedFields` from it):

```
kind: NonAdminBackup
metadata:
  creationTimestamp: '2024-05-02T08:27:46Z'
  name: empty-backupspec
  namespace: nacproject
  resourceVersion: '71261871'
  uid: 624c270e-2711-490d-923e-aaebfcc2a420
spec:
  backupSpec:
    volumeSnapshotLocations:
      - velero-sample-1
    defaultVolumesToFsBackup: false
    csiSnapshotTimeout: 10m0s
    ttl: 720h0m0s
    itemOperationTimeout: 4h0m0s
    metadata: {}
    storageLocation: velero-sample-1
    hooks: {}
    includedNamespaces:
      - nacproject
    snapshotMoveData: false
status:
  conditions:
    - lastTransitionTime: '2024-05-02T08:27:47Z'
      message: backup accepted
      reason: BackupAccepted
      status: 'True'
      type: Accepted
    - lastTransitionTime: '2024-05-02T08:27:57Z'
      message: Created Velero Backup object
      reason: BackupScheduled
      status: 'True'
      type: Queued
  phase: Created
  veleroBackupName: nab-nacproject-d0b80d478b6a26
  veleroBackupNamespace: openshift-adp
  veleroBackupStatus:
    expiration: '2024-06-01T08:27:57Z'
    failureReason: >-
      unable to get credentials: unable to get key for secret: Secret
      "cloud-credentials" not found
    formatVersion: 1.1.0
    phase: Failed
    startTimestamp: '2024-05-02T08:27:57Z'
    version: 1
```

c) Perform step a) and then update/patch (a) to include empty backupSpec. The reconcile should be same as in b), so the Velero Backup should be created.

d) Create NonAdminBackup with `includedNamespaces` that does not match the namespace from the NonAdminBackup.

```
kind: NonAdminBackup
metadata:
  name: example-wrongnamespace
  namespace: nacproject
spec:
   backupSpec:
     includedNamespaces:
       - othernamespace
```

Result should be NAB in backingOff phase with `spec.backupSpec.IncludedNamespaces can not contain namespaces other then: nacproject` condition message:

```
kind: NonAdminBackup
metadata:
  creationTimestamp: '2024-05-02T08:33:35Z'
  generation: 1
  name: example-wrongnamespace
  namespace: nacproject
  resourceVersion: '71264312'
  uid: 2716a72d-3e44-466d-8126-5bcdeed9413b
spec:
  backupSpec:
    includedNamespaces:
      - othernamespace
status:
  conditions:
    - lastTransitionTime: '2024-05-02T08:33:45Z'
      message: >-
        spec.backupSpec.IncludedNamespaces can not contain namespaces other
        then: nacproject
      reason: InvalidBackupSpec
      status: 'False'
      type: Accepted
  phase: BackingOff
```


mpryc:
need to rebase...

internal/controller/nab_init.go
mateusoliveira43, line 1: I would delete all these new files, adding code direct to nonadminbackup_controller.go

internal/controller/nonadminbackup_controller.go
```diff
@@ -70,23 +62,18 @@ const (
...
-	r.NamespacedName = req.NamespacedName
-
-	// Get the Non Admin Backup object
+	// Get the NonAdminBackup object
 	nab := nacv1alpha1.NonAdminBackup{}
 	err := r.Get(ctx, req.NamespacedName, &nab)
 
 	// Bail out when the Non Admin Backup reconcile was triggered, when the NAB got deleted
 	// Reconcile loop was triggered when Velero Backup object got updated and NAB isn't there
 	if err != nil && apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
mpryc: Fixed:

https://github.com/migtools/oadp-non-admin/pull/54/commits/c91e31a9788e167b4b88013bc3e980c1e9adfbe1

internal/controller/nonadminbackup_controller.go
```diff
@@ -70,167 +62,46 @@ const (
...
+		ReconcileFunc(func(...any) (bool, bool, error) {
+			return r.ValidateVeleroBackupSpec(ctx, rLog, &nab) // Phase: New
+		}),
+		ReconcileFunc(func(...any) (bool, bool, error) {
+			return r.CreateVeleroBackupSpec(ctx, rLog, &nab) // Phase: New
+		}),
+	)
+
+	// Return vars from the ReconcileBatch
+	if reconcileExit {
```
mateusoliveira43, line 98: would also simplify here
```go
	if reconcileErr != nil  || reconcileExit {
		return ctrl.Result{}, reconcileErr
	}
        if reconcileRequeue {
		return ctrl.Result{Requeue: true, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	}
```

internal/controller/nonadminbackup_controller.go
```diff
@@ -70,167 +62,46 @@ const (
...
-		return ctrl.Result{}, errUpdate
-	}
-	_, errUpdate = function.UpdateNonAdminBackupCondition(ctx, r.Client, logger, &nab, nacv1alpha1.NonAdminConditionQueued, metav1.ConditionTrue, "BackupScheduled", "Created Velero Backup object")
-	if errUpdate != nil {
-		logger.Error(errUpdate, "Unable to set BackupQueued Condition: True", nameField, req.Name, constant.NameSpaceString, req.Namespace)
-		return ctrl.Result{}, errUpdate
-	}
+	// Run Reconcile Batch as we have the NonAdminBackup object
+	reconcileExit, reconcileRequeue, reconcileErr := ReconcileBatch(
+		ReconcileFunc(func(...any) (bool, bool, error) {
```
mateusoliveira43, line 86: all these new functions have same params (ctx, log, nab, right :question: I would enforcing typing here

internal/controller/nonadminbackup_controller.go
mpryc: How about even simpler and entire thing replace with. See we don't really care here about reconcileExit anymore, because this is the same as the last line of the reconcile loop ?:

```
	// Run Reconcile Batch as we have the NonAdminBackup object
	_, reconcileRequeue, reconcileErr := ReconcileBatch(
		ReconcileFunc(func(...any) (bool, bool, error) {
			return r.InitNonAdminBackup(ctx, rLog, &nab) // Phase: New
		}),
		ReconcileFunc(func(...any) (bool, bool, error) {
			return r.ValidateVeleroBackupSpec(ctx, rLog, &nab) // Phase: New
		}),
		ReconcileFunc(func(...any) (bool, bool, error) {
			return r.CreateVeleroBackupSpec(ctx, rLog, &nab) // Phase: New
		}),
	)

	// Return vars from the ReconcileBatch
	if reconcileRequeue {
		return ctrl.Result{Requeue: true, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	}

	logger.V(1).Info(">>> Reconcile NonAdminBackup - loop end")
	return ctrl.Result{}, reconcileErr
}
```

internal/controller/nonadminbackup_controller.go
mpryc: They may not have the same in the future, we will be adding different batch for Restore etc... and I want this to be common function.

internal/controller/nonadminbackup_controller.go
mpryc: Adjusted last commit with the above change. Note we do have now `_` as the return from ReconcileBatch, we may want to either remove first bool or keep it depending on how other controllers will utilize that function.

mpryc: See the oadp has similar `_` implemented: https://github.com/openshift/oadp-operator/blob/master/controllers/dpa_controller.go#L93

mateusoliveira43: and it should be removed as well https://github.com/search?q=repo%3Aopenshift%2Foadp-operator+ReconcileBatch+language%3AGo&type=code&l=Go 

exit does the same job as a return (exit this function early, with or without error). It should be removed

internal/controller/nab_init.go
```diff
@@ -0,0 +1,58 @@
...
+//	ctx: Context for the request.
+//	log: Logger instance for logging messages.
+//	nab: Pointer to the NonAdminBackup object.
+//
+// The function checks if the Phase of the NonAdminBackup object is empty.
+// If it is empty, it sets the Phase to "New".
+// It then returns boolean values indicating whether the reconciliation loop should requeue
+// and whether the status was updated.
+func (r *NonAdminBackupReconciler) InitNonAdminBackup(ctx context.Context, log logr.Logger, nab *nacv1alpha1.NonAdminBackup) (exitReconcile bool, requeueReconcile bool, errorReconcile error) {
+	logger := log.WithValues("NonAdminBackup", nab.Namespace)
```
mateusoliveira43, line 42: all these functions start with this. Can not it be initialized outside the functions :question:

internal/controller/nab_init.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+
+	"github.com/go-logr/logr"
+
+	nacv1alpha1 "github.com/migtools/oadp-non-admin/api/v1alpha1"
+	"github.com/migtools/oadp-non-admin/internal/common/constant"
+	"github.com/migtools/oadp-non-admin/internal/common/function"
+)
+
+// InitNonAdminBackup Initsets the New Phase on a NonAdminBackup object if it is not already set.
```
mateusoliveira43, line 29: Initsets --> sets :question:

internal/controller/nab_init.go
```diff
@@ -0,0 +1,58 @@
...
+	logger := log.WithValues("NonAdminBackup", nab.Namespace)
+	// Set initial Phase
+	if nab.Status.Phase == constant.EmptyString {
+		// Phase: New
+		updatedStatus, errUpdate := function.UpdateNonAdminPhase(ctx, r.Client, logger, nab, nacv1alpha1.NonAdminBackupPhaseNew)
+		if updatedStatus {
+			logger.V(1).Info("NonAdminBackup CR - Rqueue after Phase Update")
+			return false, true, nil
+		}
+		if errUpdate != nil {
```
mateusoliveira43, line 51: should not error be checked first than first return argument :question:

internal/controller/nab_validator.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	nacv1alpha1 "github.com/migtools/oadp-non-admin/api/v1alpha1"
+	"github.com/migtools/oadp-non-admin/internal/common/constant"
+	"github.com/migtools/oadp-non-admin/internal/common/function"
+)
+
+// ValidateVeleroBackupSpec validates the VeleroBackup Spec from the NonAdminBackup.
```
mateusoliveira43, line 30: the VeleroBackup Spec --> spec.backupSpec :question:

internal/controller/nab_validator.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	updatedStatus, errUpdateStatus := function.UpdateNonAdminBackupCondition(ctx, r.Client, logger, nab, nacv1alpha1.NonAdminConditionAccepted, metav1.ConditionTrue, "BackupAccepted", "backup accepted")
+	if updatedStatus {
+		// We do requeue - The VeleroBackup got accepted and next reconcile loop will continue
+		// with further work on the VeleroBackup such as creating it
+		return false, true, nil
+	}
+
+	if errUpdateStatus != nil {
```
mateusoliveira43, line 84: should not check error first ❓

internal/controller/nab_validator.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if err != nil {
+		errMsg := "NonAdminBackup CR does not contain valid BackupSpec"
+		logger.Error(err, errMsg)
+
+		updatedStatus, errUpdateStatus := function.UpdateNonAdminPhase(ctx, r.Client, logger, nab, nacv1alpha1.NonAdminBackupPhaseBackingOff)
+		if errUpdateStatus != nil {
+			logger.Error(errUpdateStatus, "Unable to set NonAdminBackup Phase: BackingOff", nameField, nab.Name, constant.NameSpaceString, nab.Namespace)
+			return true, false, errUpdateStatus
+		} else if updatedStatus {
```
mateusoliveira43, line 57: this part is happy path, no errors happened, right :question: should not it set condition prior to exiting :question:

internal/controller/nab_velero_backupspec.go
```diff
@@ -0,0 +1,136 @@
...
+				Name:      veleroBackupName,
+				Namespace: constant.OadpNamespace,
+			},
+			Spec: *backupSpec,
+		}
+	} else if err != nil && !apierrors.IsNotFound(err) {
+		logger.Error(err, "Unable to fetch VeleroBackup")
+		return true, false, err
+	} else {
+		// We should not update already created VeleroBackup object.
```
mateusoliveira43, line 84: here, it should not generate a new name and try again :question:

internal/controller/nab_init.go
mateusoliveira43, line 1: no tests for these functions, we should add, right :question:

internal/controller/nab_validator.go
mpryc: No, the happy path is the code after 57 line this is semi-happy path where the status was updated, but the reconcile should continue with different path if the status did not require update.

internal/controller/nab_validator.go
mpryc: No, that is also this semi-happy path where we had `updatedStatus`. We re-queue here and not exit from reconcile.

Exit reconcile happens at the end of that function which is different from re-queue.

internal/controller/nab_validator.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		// Continue. VeleroBackup looks fine, setting Accepted condition
+		updatedCondition, errUpdateCondition := function.UpdateNonAdminBackupCondition(ctx, r.Client, logger, nab, nacv1alpha1.NonAdminConditionAccepted, metav1.ConditionFalse, "InvalidBackupSpec", errMsg)
+		if updatedCondition {
+			// We do not requeue - this was only Condition update
+			return true, false, nil
+		}
+
+		if errUpdateCondition != nil {
```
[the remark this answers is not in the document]
mpryc: same comment as on the line 84.

internal/controller/nab_validator.go
mpryc: isn't that the same as `VeleroBackup Spec from ... ` I wanted to ensure people understand this is VeleroBackup specification and not some backupSpec which isn't clearly visible that is type of VeleroBackup

internal/controller/nab_init.go
mpryc: no as in other comments we have 3 states here:
 - correct
 - correct with requeue (update status)
 - error

internal/controller/nab_init.go
mpryc: yep, typo.

internal/controller/nab_init.go
shawn-hurley: It probably shouldn't be outside the functions, this is the logger for the thing that is being inited IIUC.

internal/controller/nab_init.go
shawn-hurley: These are just my two cents. I think you are saying that the states are mutually exclusive. There is only ever,

false, nil (correct)
true, nil (requeue)
false, err (error)

If that is the case, I would suggest that handling the third case first will look more go idiomatic for the future.

If there is a fourth option:
true, err 

If this is the case, an error will be required, but the status will not be updated IIUC. I think it makes sense if this is an important case, then we handle the request first so that we update the status.

internal/controller/nac_reconcile_utils.go
```diff
@@ -0,0 +1,81 @@
...
+// ReconcileFunc represents a function that performs a reconciliation operation.
+// Parameters:
+//   - ...any: A variadic number of arguments,
+//     allowing for flexibility in passing parameters to the reconciliation function.
+//
+// Returns:
+//   - bool: Indicates whether the reconciliation process should exit.
+//   - bool: Indicates whether the reconciliation process should requeue.
+//   - error: An error encountered during reconciliation, if any.
+type ReconcileFunc func(...any) (bool, bool, error)
```
shawn-hurley, line 30: I don't know if we should just remove the type system; I would probably caution against this; I would much rather have a long function that calls individual functions but has the type safety.

I also don't see any of the parameters are ever passed in here.

mpryc: We could simply have, then entire type and Batch is gone:
```

	reconcileExit, reconcileRequeue, reconcileErr := r.InitNonAdminBackup(ctx, rLog, &nab)
	if reconcileRequeue {
		return ctrl.Result{Requeue: true, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	} else if reconcileExit || reconcileErr != nil {
		return ctrl.Result{}, reconcileErr
	}

	_, reconcileRequeue, reconcileErr = r.ValidateVeleroBackupSpec(ctx, rLog, &nab)
	if reconcileRequeue {
		return ctrl.Result{Requeue: true, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	} else if reconcileExit || reconcileErr != nil {
		return ctrl.Result{}, reconcileErr
	}

	_, reconcileRequeue, reconcileErr = r.CreateVeleroBackupSpec(ctx, rLog, &nab)
	if reconcileRequeue {
		return ctrl.Result{Requeue: true, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	} else if reconcileExit || reconcileErr != nil {
		return ctrl.Result{}, reconcileErr
	}
```

internal/controller/nab_init.go
```diff
@@ -0,0 +1,58 @@
...
+// It then returns boolean values indicating whether the reconciliation loop should requeue
+// and whether the status was updated.
+func (r *NonAdminBackupReconciler) InitNonAdminBackup(ctx context.Context, log logr.Logger, nab *nacv1alpha1.NonAdminBackup) (exitReconcile bool, requeueReconcile bool, errorReconcile error) {
+	logger := log.WithValues("NonAdminBackup", nab.Namespace)
+	// Set initial Phase
+	if nab.Status.Phase == constant.EmptyString {
+		// Phase: New
+		updatedStatus, errUpdate := function.UpdateNonAdminPhase(ctx, r.Client, logger, nab, nacv1alpha1.NonAdminBackupPhaseNew)
+		if updatedStatus {
+			logger.V(1).Info("NonAdminBackup CR - Rqueue after Phase Update")
```
PrasadJoshi12, line 48:
```suggestion
			logger.V(1).Info("NonAdminBackup CR - Requeue after Phase Update")
```

internal/controller/nonadminbackup_controller.go
mpryc: Ok I have very much reworked that part and this comment is not relevant anymore with the new commit which is not using ReconcileBatch and we need args there.

internal/controller/nab_init.go
mpryc: Corrected log to include function, comment not relevant, resolving. Please re-open if you see additional improvements in the new commit.

internal/controller/nab_init.go
mpryc: Updated to first check for errors. This was more of a 3rd state, which I checked can be safely moved to first check for the error. Closing as resolved as this is reworked in new commit. Please re-open if new implementation is also incorrect in your opinion.

internal/controller/nab_init.go
mpryc: Thanks. Fixed.

internal/controller/nonadminbackup_controller.go
mpryc: Current implementation is very different without ReconcileBatch so we don't have that anymore.

internal/controller/nab_velero_backupspec.go
mateusoliveira43: no no, I meant: is not this triggered when the generated velero backup name already exists :question: which can be considered corner case...

internal/controller/nonadminbackup_controller.go
```diff
@@ -70,170 +62,47 @@ const (
...
-	if errUpdate != nil {
-		logger.Error(errUpdate, "Unable to set BackupAccepted Condition: True", nameField, req.Name, constant.NameSpaceString, req.Namespace)
-		return ctrl.Result{}, errUpdate
-	}
-	_, errUpdate = function.UpdateNonAdminBackupCondition(ctx, r.Client, logger, &nab, nacv1alpha1.NonAdminConditionQueued, metav1.ConditionTrue, "BackupScheduled", "Created Velero Backup object")
-	if errUpdate != nil {
-		logger.Error(errUpdate, "Unable to set BackupQueued Condition: True", nameField, req.Name, constant.NameSpaceString, req.Namespace)
-		return ctrl.Result{}, errUpdate
+	reconcileExit, reconcileRequeue, reconcileErr = r.CreateVeleroBackupSpec(ctx, rLog, &nab)
+	if reconcileRequeue {
```
mateusoliveira43, line 99: I think @shawn-hurley commented in another thread: how many options do we have :question: if it is 3 ([false,err],[false,nil] and [true,nil] for re queue and error) we do not need `reconcileExit` on these functions, correct :question: 

and would check `reconcileErr` prior to re queue

internal/controller/nab_init.go
shubham-pampattiwar: you can keep the validator file as seperate but merge the other 2 in `nonadminbackup_controller.go`

internal/controller/nab_velero_backupspec.go
shubham-pampattiwar: I think that case is already covered in the line 58 conditional statement, right ?

internal/controller/nonadminbackup_controller.go
mpryc: We have the following logic that needs to be propagated back to main Reconcile function:

 1. Exit main reconcile immediately (true, false, nil)
 2. Exit main reconcile immediately with error (true, false, error)
 3. Re-queue main reconcile (false, true, nil)
 4. Continue with the main logic (false, false, nil)

We need those 4 cases because the main reconcile may continue (4th case).

Those 4 cases from which 3 have no error can not be written using one true/false return which gives us only 2 cases.

Even if we decide to include error as the logic part I still want to have explicit flag that tells if the main reconcile is re-queued, exited or continued. The logic is already complex and for the readability we definitely want those flags to be  clearly stating what is the path of the reconcile.

internal/controller/nab_velero_backupspec.go
mpryc: Fixed. Mistake during move of the code logic. This is corner case anyway.

I am wondering why linting didn't catch this as it's same logic in two if/else places.

internal/controller/nab_validator.go
mpryc: No. We do not re-queue on status or condition update. The 67 did modify condition already and the reconcile loop got status update so we reconciled again, meaning this is second time this function is called.

That is why we have `updatedStatus` flag to ensure we don't go into infinite loop of reconciles and only update when needed.

The status update and condition update are separate and independent this way and it's better as they may be populated in different workers / times.

internal/controller/nab_init.go
mpryc: I am in favor of merging all of them or none of them, to be consistent.

mpryc:
> no "how to test" section in PR description

Added with few different use-cases.

internal/controller/nab_init.go
mateusoliveira43: I think we need discussion for E2E (no need for those now), unit and integration should be just use already created structure, no :question: 

example of integration tests I wrote https://github.com/migtools/oadp-non-admin/pull/42/files#diff-01fb7d70bad3304acc2b09564e508353cd0a83094b735dad6e8a153bdff85f80

shubham-pampattiwar: ok, you can merge all of them then

mpryc: @shubham-pampattiwar merge them all and move to nonadminbackup_controller.go or merge them all and move them to separate file kind of `nonadminbackup_reconcile_functions.go` ?

shubham-pampattiwar: we can merge them in `nonadminbackup_controller.go`

mpryc: done.

internal/controller/nonadminbackup_controller.go
mateusoliveira43: Alright, what about keeping as it is, but simplify `if`s :question: 
```go
	if reconcileExit || reconcileRequeue || reconcileErr != nil{
		return ctrl.Result{Requeue: reconcileRequeue, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	}
```

internal/controller/nonadminbackup_controller.go
```diff
@@ -70,101 +67,180 @@ const (
...
-			return ctrl.Result{}, nil
+			// We do not requeue - the State was set to BackingOff
+			return true, false, nil
 		}
 
-		updatedStatus, errUpdate = function.UpdateNonAdminBackupCondition(ctx, r.Client, logger, &nab, nacv1alpha1.NonAdminConditionAccepted, metav1.ConditionFalse, "InvalidBackupSpec", errMsg)
-		if updatedStatus {
-			return ctrl.Result{}, nil
-		}
+		// Continue. VeleroBackup looks fine, setting Accepted condition
```
shubham-pampattiwar, line 179: I am assuming the NAB spec validation logic will reside here in between setting `BackupAccepted` as `False` and then setting `BackupAccepted` as `True`

mateusoliveira43: I think would be here https://github.com/mpryc/oadp-non-admin/blob/c1983f9f1ec3c603a8ad47f6deb7b0baf05ed66f/internal/common/function/function.go#L86

but that function validates and gets spec

I think would be better to split in 2 funtions

shubham-pampattiwar: yes splitting would make sense, I agree +1

shubham-pampattiwar:
Did some high level testing, 2 issues that can be addressed separately in follow up PRs:
- https://github.com/migtools/oadp-non-admin/issues/58
- https://github.com/migtools/oadp-non-admin/issues/59
Rest looks good, Thank you @mpryc

internal/controller/nonadminbackup_controller.go
mpryc: Can we have this as follow up PR? It's easier to test small change then such bigger one.

shubham-pampattiwar: yeah follow up would work, created an issue to track it https://github.com/migtools/oadp-non-admin/issues/60

shubham-pampattiwar (approved):

internal/controller/nonadminbackup_controller.go
mpryc: The two implementations are not the same, plus your hides the desired logic inside  merged logical expression which in my opinion is harder to follow.

internal/controller/nab_init.go
mpryc: They are, but we can use many ways to perform tests (against live cluster vs mocked calls) so I would like to discuss those on separate PR.

internal/controller/nonadminbackup_controller.go
mpryc: @mateusoliveira43 Can we merge with the current implementation? If you think it's wrong PRs are welcome.

internal/controller/nab_init.go
mpryc: https://github.com/migtools/oadp-non-admin/issues/61

mpryc: Closing comment as this is tracked via 61

internal/controller/nonadminbackup_controller.go
mpryc: Tracked via https://github.com/migtools/oadp-non-admin/issues/62 . Closing comment.

Issue #49, opened 2024-04-23
Ensure non-admin backups are namespace scoped


weshayutin:
With a valid namespace set and credentials a non-admin user is able to take a backup of cr's outside the scope of the user's namespace(s).

```
apiVersion: nac.oadp.openshift.io/v1alpha1
kind: NonAdminBackup
metadata:
  name: empty-backup-spec
  namespace: mysql-persistent
spec:
  backupSpec: {}
```

results in:

```
status:
  conditions:
  - lastTransitionTime: "2024-04-23T18:48:23Z"
    message: backup accepted
    reason: BackupAccepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-04-23T18:48:33Z"
    message: Created Velero Backup object
    reason: BackupScheduled
    status: "True"
    type: Queued
  phase: Created
  veleroBackupName: nab-mysql-persistent-e33e97f12f2303
  veleroBackupNamespace: openshift-adp
  veleroBackupStatus:
    backupItemOperationsAttempted: 62
    backupItemOperationsCompleted: 60
    csiVolumeSnapshotsAttempted: 31
    csiVolumeSnapshotsCompleted: 24
    expiration: "2024-05-23T18:48:33Z"
    formatVersion: 1.1.0
    phase: WaitingForPluginOperations
    progress:
      itemsBackedUp: 8241
      totalItems: 8241
    startTimestamp: "2024-04-23T18:48:33Z"
    version: 1

```

A properly namespaced backup should have only 83 items in this case.
```
status:
  conditions:
  - lastTransitionTime: "2024-04-23T18:38:54Z"
    message: backup accepted
    reason: BackupAccepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-04-23T18:39:04Z"
    message: Created Velero Backup object
    reason: BackupScheduled
    status: "True"
    type: Queued
  phase: Created
  veleroBackupName: nab-mysql-persistent-a0a611237fcdb4
  veleroBackupNamespace: openshift-adp
  veleroBackupStatus:
    backupItemOperationsAttempted: 2
    backupItemOperationsCompleted: 2
    completionTimestamp: "2024-04-23T18:39:17Z"
    csiVolumeSnapshotsAttempted: 1
    csiVolumeSnapshotsCompleted: 1
    expiration: "2024-05-23T18:39:04Z"
    formatVersion: 1.1.0
    phase: Completed
    progress:
      itemsBackedUp: 83
[...]

PR #56, opened 2024-04-30
Validate namespaces, so non admin can not create backups outside of ns

mpryc:
Validate namespaces, so non admin can not create backups outside of ns
    
Validation ensures the namespace in the VeleroBackup object are the same as the namespace for which NAB resides.

Note this PR is on top of #54 which should be merged first.

# How to test

1. Setup non-admin user as per instructions: https://github.com/migtools/oadp-non-admin/blob/master/docs/non_admin_user.md
2. Create two NonAdminBackup objects in the namespace for which the non-admin user have namespace admin rights. After each creation see if at the end of reconcile the NonAdminBackup is correctly updated with its status and if in some cases VeleroBackup object was created in the openshift-adp namespace (as an admin) or not created if there was issue with namespace.

All of the below are in the non-admin namespace `nacproject`:

a) Case where there is backupSpec, but empty, should create Velero Backup with nacproject namespace in the includedNamespaces (below is not full spec, removed `managedFields` from it):

```
kind: NonAdminBackup
metadata:
  name: empty-backupspec
  namespace: nacproject
spec: 
  backupSpec: {}
```

Result should be created Velero Backup with `includedNamespaces` same as the origin NAB namespace and updated NonAdminBackup (below is not full spec, removed `managedFields` from it):

```
kind: NonAdminBackup
metadata:
  creationTimestamp: '2024-05-02T08:27:46Z'
  name: empty-backupspec
  namespace: nacproject
  resourceVersion: '71261871'
  uid: 624c270e-2711-490d-923e-aaebfcc2a420
spec:
  backupSpec:
    volumeSnapshotLocations:
      - velero-sample-1
    defaultVolumesToFsBackup: false
    csiSnapshotTimeout: 10m0s
    ttl: 720h0m0s
    itemOperationTimeout: 4h0m0s
    metadata: {}
    storageLocation: velero-sample-1
    hooks: {}
    includedNamespaces:
      - nacproject
    snapshotMoveData: false
status:
  conditions:
    - lastTransitionTime: '2024-05-02T08:27:47Z'
      message: backup accepted
      reason: BackupAccepted
      status: 'True'
      type: Accepted
    - lastTransitionTime: '2024-05-02T08:27:57Z'
      message: Created Velero Backup object
      reason: BackupScheduled
      status: 'True'
      type: Queued
  phase: Created
  veleroBackupName: nab-nacproject-d0b80d478b6a26
  veleroBackupNamespace: openshift-adp
  veleroBackupStatus:
    expiration: '2024-06-01T08:27:57Z'
    failureReason: >-
      unable to get credentials: unable to get key for secret: Secret
      "cloud-credentials" not found
    formatVersion: 1.1.0
    phase: Failed
    startTimestamp: '2024-05-02T08:27:57Z'
    version: 1
```

b) Create NonAdminBackup with `includedNamespaces` that does not match the namespace from the NonAdminBackup. It may include namespace of the NAB object, but on top of that it should have also other namespaces.

```
kind: NonAdminBackup
metadata:
  name: example-wrongnamespace
  namespace: nacproject
spec:
   backupSpec:
     includedNamespaces:
       - othernamespace
       - yetanothernamespace
       - nacproject
```

Result should be NAB in backingOff phase with `spec.backupSpec.IncludedNamespaces can not contain namespaces other then: nacproject` condition message:

```
kind: NonAdminBackup
metadata:
  creationTimestamp: '2024-05-02T08:33:35Z'
  generation: 1
  name: example-wrongnamespace
  namespace: nacproject
  resourceVersion: '71264312'
  uid: 2716a72d-3e44-466d-8126-5bcdeed9413b
spec:
  backupSpec:
    includedNamespaces:
       - othernamespace
       - yetanothernamespace
       - nacproject
status:
  conditions:
    - lastTransitionTime: '2024-05-02T08:33:45Z'
      message: >-
        spec.backupSpec.IncludedNamespaces can not contain namespaces other
        then: nacproject
      reason: InvalidBackupSpec
      status: 'False'
      type: Accepted
  phase: BackingOff
```

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -77,21 +77,39 @@ func AddNonAdminBackupAnnotations(ownerNamespace string, ownerName string, owner
 	return mergedAnnotations
 }
 
+// containsOnlyNamespace checks if the given namespaces slice contains only the specified namespace
+func containsOnlyNamespace(namespaces []string, namespace string) bool {
+	for _, ns := range namespaces {
+		if ns != namespace {
+			return false
+		}
+	}
+	return true
+}
+
 // GetBackupSpecFromNonAdminBackup return BackupSpec object from NonAdminBackup spec, if no error occurs
 func GetBackupSpecFromNonAdminBackup(nonAdminBackup *nacv1alpha1.NonAdminBackup) (*velerov1api.BackupSpec, error) {
 	if nonAdminBackup == nil {
 		return nil, fmt.Errorf("nonAdminBackup is nil")
 	}
 
-	// TODO check spec?
-
 	if nonAdminBackup.Spec.BackupSpec == nil {
-		return nil, fmt.Errorf("BackupSpec is nil")
+		return nil, fmt.Errorf("BackupSpec is not defined")
 	}
+
+	veleroBackupSpec := nonAdminBackup.Spec.BackupSpec.DeepCopy()
 
 	// TODO: Additional validations, before continuing
 
-	return nonAdminBackup.Spec.BackupSpec.DeepCopy(), nil
+	if veleroBackupSpec.IncludedNamespaces == nil {
+		veleroBackupSpec.IncludedNamespaces = []string{nonAdminBackup.Namespace}
+	} else {
+		if !containsOnlyNamespace(veleroBackupSpec.IncludedNamespaces, nonAdminBackup.Namespace) {
+			return nil, fmt.Errorf("spec.backupSpec.IncludedNamespaces can not contain namespaces other then: %s", nonAdminBackup.Namespace)
+		}
+	}
+
+	return veleroBackupSpec, nil
 }
 
 // GenerateVeleroBackupName generates a Velero backup name based on the provided namespace and NonAdminBackup name.
@@ -185,7 +203,7 @@ func UpdateNonAdminBackupCondition(ctx context.Context, r client.Client, logger 
 
 	// Check if the condition is already set to the desired status
 	currentCondition := apimeta.FindStatusCondition(nab.Status.Conditions, string(condition))
-	if currentCondition != nil && currentCondition.Status == conditionStatus {
+	if currentCondition != nil && currentCondition.Status == conditionStatus && currentCondition.Reason == reason && currentCondition.Message == message {
 		// Condition is already set to the desired status, no need to update
 		logger.V(1).Info(fmt.Sprintf("NonAdminBackup Condition is already set to: %s", condition))
 		return false, nil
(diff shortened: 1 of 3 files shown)


mpryc:
@mateusoliveira43 @shubham-pampattiwar @weshayutin  This PR is really all about that commit only, because other commits are covered in #54:

https://github.com/migtools/oadp-non-admin/pull/56/commits/152d1c7ba3466c37a66fff2304f111136182e351

mpryc:
Fixes #49

internal/common/function/function.go
```diff
@@ -77,21 +77,39 @@ func AddNonAdminBackupAnnotations(ownerNamespace string, ownerName string, owner
...
+		return nil, fmt.Errorf("BackupSpec is not defined")
 	}
 
+	veleroBackupSpec := nonAdminBackup.Spec.BackupSpec.DeepCopy()
+
 	// TODO: Additional validations, before continuing
 
-	return nonAdminBackup.Spec.BackupSpec.DeepCopy(), nil
+	if veleroBackupSpec.IncludedNamespaces == nil {
+		veleroBackupSpec.IncludedNamespaces = []string{nonAdminBackup.Namespace}
```
mateusoliveira43, line 105: does this change NonAdminBackup spec or just passes a different one to velero backup that will be created :question:

internal/common/function/function_test.go
```diff
@@ -196,7 +198,10 @@ func TestGetBackupSpecFromNonAdminBackup(t *testing.T) {
 
 	assert.NoError(t, err)
 	assert.NotNil(t, backupSpec)
-	assert.Equal(t, backupSpecInput, backupSpec)
+	assert.Equal(t, []string{"namespace1"}, backupSpec.IncludedNamespaces) // Check that only the namespace from NonAdminBackup is included
+	assert.Equal(t, backupSpecInput.ExcludedNamespaces, backupSpec.ExcludedNamespaces)
+	assert.Equal(t, backupSpecInput.StorageLocation, backupSpec.StorageLocation)
+	assert.Equal(t, backupSpecInput.VolumeSnapshotLocations, backupSpec.VolumeSnapshotLocations)
 }
```
mateusoliveira43, line 205: add test for new error :question:

internal/common/function/function.go
mpryc: So when there is no includenamespaces then it's changed otherwise it sets condition accepted to false.

internal/common/function/function_test.go
mpryc: ah right there is missing one. will do

mpryc: Added 2 new test cases to cover new error.

mpryc:
Rebased validation of namespaces. @mateusoliveira43 @shubham-pampattiwar please review as it's on top of previous work with merged reconcile functions.

shubham-pampattiwar (approved):
/lgtm

Issue #65, opened 2024-05-13
Reconcile loop is not exitting properly


mpryc:
Changed introduced after `94c4be9bf93da2d791281d2aaca54655d28213161` caused reconcile loop to not exit properly and when the state is incorrect e.g. `BackingOff` re-enter reconcile.

mateusoliveira43:
did not found this commit, can you link it?

mpryc:
The last correctly working was:

`fix: Add more documenation to project (#16)` One of later commits caused the behavior.

To reproduce:

Create NonAdminBackup without proper spec, e.g. Then reconcile loop will enter every few sec.

```
apiVersion: nac.oadp.openshift.io/v1alpha1
kind: NonAdminBackup
metadata:
  name: example
  namespace: nacproject
spec: {}
```

mpryc:
@mateusoliveira43 The problem is with the following line. If error is returned then the reconcile is re-entered:

https://github.com/migtools/oadp-non-admin/blob/master/internal/controller/nonadminbackup_controller.go#L100

mpryc:
What we should do here is to:

```
1. Have the Error exposed via:
logger.Error()

2. Return nil as an error
```

The above will show the error in the controller but will not re-enter the reconcile. We should only re-enter reconcile if we requeue it.

mateusoliveira43:
maybe err in reconcile is meant, for example, http err and then retry reconcile?

log the error seems ok, but would say priority is to add message to CRD status, right?

so we would re run reconcile if:

1. re queue is needed
2. unexpected err occurred

for "normal" errors, like wrong user spec in NAB, reconcile exits nil, but CRD status contains error

That seems correct @shubham-pampattiwar @shawn-hurley  ?

shawn-hurley:
@mateusoliveira43 If there is no reason to reconcile, because we know that the error is needed to be handled by the end user, it would be smart to not re-queue. I believe the way to implement this behavior is [terminal errors](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#TerminalError). 

IIRC there is a field on the reconcile `Result` object that handles [requeueing](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result).

mpryc:
@shawn-hurley thanks I was looknig at terminal errors as well. The requeue false in the Result is skipped if an error is returned.

mateusoliveira43:
@mpryc is this done?

[...]

PR #66, opened 2024-05-14
Fix for the Reconcile loop is not exitting properly

mpryc:
Fixes issue #65

Whenever we want to exit with error state from the reconcile loop that is considered final state we should use TerminalError.

Returning standard error will cause Reconcile to requeue.

## Testing
1. Deploy controller without the change
2. Create NonAdminBackup without proper backupSpec, for example
```yaml
kind: NonAdminBackup
metadata:
  name: example
  namespace: nacproject
spec: {}
```

3. Observe controller log. It will show error, however it will re-enter check every few seconds.

With this fix, the point number 3 will not re-enter check every few seconds.

In either scenario proper NonAdminBackup object phase and conditions are correct.

### DIFF
--- a/internal/controller/nonadminbackup_controller.go
+++ b/internal/controller/nonadminbackup_controller.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	nacv1alpha1 "github.com/migtools/oadp-non-admin/api/v1alpha1"
 	"github.com/migtools/oadp-non-admin/internal/common/constant"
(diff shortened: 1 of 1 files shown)


internal/controller/nonadminbackup_controller.go
```diff
@@ -89,22 +90,28 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	reconcileExit, reconcileRequeue, reconcileErr := r.InitNonAdminBackup(ctx, rLog, &nab)
 	if reconcileRequeue {
 		return ctrl.Result{Requeue: true, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
-	} else if reconcileExit || reconcileErr != nil {
-		return ctrl.Result{}, reconcileErr
+	} else if reconcileExit && reconcileErr != nil {
+		return ctrl.Result{}, reconcile.TerminalError(reconcileErr)
```
mateusoliveira43, line 94: instead of adding another else here, can `reconcile.TerminalError` be applied in functions? For example, inside `InitNonAdminBackup` function

internal/controller/nonadminbackup_controller.go
mateusoliveira43, line 1: need terminal error also in here?
https://github.com/migtools/oadp-non-admin/blob/b521e5c2a410eac39dd2898c286f4c1181b3e318/internal/controller/nonadminbackup_controller.go#L86

internal/controller/nonadminbackup_controller.go
mateusoliveira43: from what I read (did not test), I think it can be applied to `nil` as well, so would not change else structure here, just wrap error in `reconcile.TerminalError`

internal/controller/nonadminbackup_controller.go
mpryc: no we don't want to always have terminal error. This is unknown to us problem, possibly some OpenShift one?

internal/controller/nonadminbackup_controller.go
mpryc: I've tested this and adding nil won't work as this will always produce ERROR in reconcile, so it's a no-go.

shubham-pampattiwar (approved):

PR #67, opened 2024-05-14
Delete Velero Backup when user sets the deleteVeleroBackup

### LINKED ISSUE #58: Facilitate deletion of VeleroBackup object via NAB object

mpryc:
Enhancement which will delete original VeleroBackup when the user sets the deleteVeleroBackup witin NonAdminBackup Spec.

Fixes #58

This allows NonAdminBackup object to have additional field within spec which will trigger deletion of the original VeleroBackup object:

```yaml
kind: NonAdminBackup
metadata:
  name: example
  namespace: nacproject
spec:
  deleteVeleroBackup: true # <<<<<< NEW 
  [...]
```

## To test:
-------
Test 1:
deleteVeleroBackup on non existing Velero Backup object
-------
1. Create NonAdminBackup without properspec

```yaml
kind: NonAdminBackup
metadata:
  name: backupwithoutspec
  namespace: nacproject
spec: {}
```

2. Ensure NonAdminBackup is in:
  - Phase:
    BackingOff
  - Has the Condition:
      Type: Accepted
      Status: False
      Reason: InvalidBackupSpec
      Message: BackupSpec is not defined

3. Update the NonAdminBackup object to include:
```yaml
[...]
spec:
  deleteVeleroBackup: true
[...]
```

4. Expected result:
   - NonAdminBackup object should be deleted by the controller.
   - Reconcile loop should exit with the following DEBUG log messages:
     DEBUG  Deleted NonAdminBackup: backupwithoutspec
     DEBUG  NonAdminBackupPredicate: Accepted Delete event
     DEBUG  >>> Reconcile NonAdminBackup - loop start
     DEBUG  Non existing NonAdminBackup CR

-------
Test 2:
deleteVeleroBackup on existing Velero Backup object set to true
-------
1. Create NonAdminBackup without properspec

```yaml
kind: NonAdminBackup
metadata:
  name: backupwithspec
  namespace: nacproject
spec:
  backupSpec: {}
```

2. Ensure NonAdminBackup is in:
  - Phase:
    Created
  - Has the Conditions:
    - Type: Accepted
      Status: True
      Reason: BackupAccepted
      Message: Backup accepted
    - Type: Accepted
      Status: True
      Reason: BackupAccepted
      Message: Backup accepted
      
3. Ensure VeleroBackup is created in the openshift-adp namespace:
   $ oc get Backup -n openshift-adp nab-nacproject-eeaa9cd35c77f4
   NAME                            AGE
   nab-nacproject-eeaa9cd35c77f4   12s

4. Update the NonAdminBackup object to include:
```yaml
[...]
spec:
  deleteVeleroBackup: true
[...]
```

5. Expected result:
   - VeleroBackup should be deleted by the controller
   - NonAdminBackup object should be deleted by the controller.
   - Reconcile loop should exit with the following DEBUG log messages:
     DEBUG  Deleted NonAdminBackup: backupwithoutspec
     DEBUG  NonAdminBackupPredicate: Accepted Delete event
     DEBUG  >>> Reconcile NonAdminBackup - loop start
     DEBUG  Non existing NonAdminBackup CR

-------
Test 3:
deleteVeleroBackup on existing Velero Backup object set to false
-------
1. Perform the same steps as 1,2,3 from the Test 2
2. Update the NonAdminBackup object to include:
```yaml
[...]
spec:
  deleteVeleroBackup: false
[...]
```
3. Reconcile loop should have the DEBUG logs:
   DEBUG: NonAdminBackup BackupSpec and BackupStatus - nothing to update

-------
Test 4:
deleteVeleroBackup on NON existing Velero Backup object set to true
-------
1. Perform the same steps as 1,2,3 from the Test 2
2. Delete Velero Backup object:
  $ oc delete -n openshift-adp nab-nacproject-eeaa9cd35c77f4
3. Update the NonAdminBackup object to include:
```yaml
[...]
spec:
  deleteVeleroBackup: true
[...]
```
4. Expected result:
   - NonAdminBackup object should be deleted by the controller.
   - Reconcile loop should exit with the following DEBUG log messages:
     DEBUG  Deleted NonAdminBackup: backupwithoutspec
     DEBUG  NonAdminBackupPredicate: Accepted Delete event
     DEBUG  >>> Reconcile NonAdminBackup - loop start
     DEBUG  Non existing NonAdminBackup CR


mpryc:
Depends on #65 as it's rebased on it.

mateusoliveira43:
@mpryc I am confused, how this fixes related issue?

This will delete Velero Backup on demand, right? But NAB will be still around

Should not be a mechanism that when NAB is deleted, Velero Backup is also deleted?

mpryc:
> @mpryc I am confused, how this fixes related issue?
> 
> This will delete Velero Backup on demand, right? But NAB will be still around
> 
> Should not be a mechanism that when NAB is deleted, Velero Backup is also deleted?

No, we discussed this with @shubham-pampattiwar and @shawn-hurley. We really can't do the above as we don't want end up in situation where there is namespace removed that triggers the NAB to be deleted and we also loose Bacups from which namespace could be restored. In k8s there isn't really separation of the events that triggered NAB deletion.

shubham-pampattiwar:
@mpryc Thoughts on why we made this feature enablement via `spec` and not via `annotation` ?

internal/common/function/function.go
```diff
@@ -280,6 +281,29 @@ func CheckVeleroBackupLabels(backup *velerov1api.Backup) bool {
 	return exists && value == constant.ManagedByLabelValue
 }
 
+// DeleteVeleroBackup deletes Velero Backup object based on the NonAdminBackup object name and namespace
+func DeleteVeleroBackup(ctx context.Context, r client.Client, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
```
shubham-pampattiwar, line 285: Isn't the Deletion logic flawed here ? Why are we using `GenerateVeleroBackupName` function to fetch the backup to be deleted ? Isnt there a high possibility that the Generated Name won't match the actual VeleroBackup Name that we want to delete ? Shouldn't we List the Velero backups and then check the annotations of these Velero backup objects and match the nab name with the annotation value of the key `"openshift.io/oadp-nab-origin-name"` to get the VeleroBackup to be deleted ? Let me know if I am missing something ?

internal/controller/nonadminbackup_controller.go
```diff
@@ -86,31 +87,91 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	reconcileExit, reconcileErr := r.DeleteVeleroBackup(ctx, rLog, &nab)
```
shubham-pampattiwar, line 90: Won't we have a case to requeue here ?

mpryc: Lint job complained that the requeue was not used in the r.DeleteVeleroBackup function, so I removed it from the logic.

internal/common/function/function.go
mpryc: Our velero backup name is *always* generated based on https://github.com/migtools/oadp-non-admin/blob/master/internal/common/function/function.go#L115-L145

```
fmt.Sprintf("%s-%s-%s", constant.VeleroBackupNamePrefix, namespace, nameHash)
```

So we have always `nab-NAMESPACE-HASH` where the HASH is sha256 from NonAdminBackup name.

VeleroBackup could get deleted only if one matches that entire 3 parts.

If we want to add extra protection we could, but then I would argue the UUID of the NonAdminBackup needs to be used from the VeleroBackup to match the UUID of the requestor.

It is currently possible, because we store this annotation within VeleroBackup:

```
openshift.io/oadp-nab-origin-uuid=<UUID_OF_NAB>
```

Let me know if I should add that protection on top of current name match.

mateusoliveira43: why not to just use 
https://github.com/migtools/oadp-non-admin/blob/43812bec2fbf7f790b9afa101c9bf395541571ff/api/v1alpha1/nonadminbackup_types.go#L52
?

shubham-pampattiwar: +1

mpryc: I thought about that, but with the current implementation I can:

 1. Create NAB
 2. VeleroBackup will get created
 3. Remove NAB
 
At this stage I have still can remove VeleroBackup from 2 if I create new NAB with the same name as the previous one with flag `deleteVeleroBackup: true`.

Using status will disallow me to do so, because 2 will never get created as this object will be there *forever* until sysadmin removes it. It's because Status can only be updated by controller when 2 is created.

shawn-hurley: We need a status condition or clear message about this backup when it gets re-attached like this: it did not run a new backup. 

I think this is the most correct behavior, but it could be surprising for an end user, and we probably need to spend time on how to make this clear to the user.

mateusoliveira43:
/hold
rebase on top of master (so PR will get smaller)
and rework deletion process as discussed in meeting Wednesday

mpryc:
Rebase and small update to the PR to include design.

@mateusoliveira43 if you could check the design update and let me know if this is fine.

To be done before is ready to merge:
 - removal of NonAdminBackup object after VeleroBackup is removed.

mateusoliveira43:
@mpryc I though we would update design here https://github.com/migtools/oadp-non-admin/pull/70. Could you push these changes there? 

But can be here and I rebase there afterwards as well

(the design changes look correct to me)

mpryc:
> @mpryc I though we would update design here #70. Could you push these changes there?
> 
> But can be here and I rebase there afterwards as well
> 
> (the design changes look correct to me)

Sure will move that update there, but now I wonder if we should completely drop status updates while deleting?

Now the way it currently works is:
 1 -> Adds Condition Deletion to True
 2 -> Remove VeleroBackup
 3 ->  Once main Velero Backup is deleted adds phase: Deleted
 4 -> Additions after Mateus idea: Removes NonAdminBackup
Now the step 3 requires requeue of reconcile loop, otherwise there is an error that the object can not be updated because it's not existing
Should we completely drop 1 and 3, because we are now removing NonAdminBackup and we don't care about phase of NAB ? What do you think ?

shubham-pampattiwar:
We need some sort of status update on NAB object once VeleroBackup is deleted because if there is an err for VeleroBackup deletion call we should be able to tell about that to the user via NAB status.

mpryc:
> We need some sort of status update on NAB object once VeleroBackup is deleted because if there is an err for VeleroBackup deletion call we should be able to tell about that to the user via NAB status.

If there is any error with VeleroBackup deletion, I think we should not try to delete NAB and only then update NAB status with:
Phase: BackingOff
Condition: Deletion -> with error what went wrong ?

shubham-pampattiwar:
> > We need some sort of status update on NAB object once VeleroBackup is deleted because if there is an err for VeleroBackup deletion call we should be able to tell about that to the user via NAB status.
> 
> If there is any error with VeleroBackup deletion, I think we should not try to delete NAB and only then update NAB status with: Phase: BackingOff Condition: Deletion -> with error what went wrong ?

yes correct that's exactly what I meant, deletion of NAB should be stopped and only status should be updated for NAB object.

mpryc:
Updated the logic to delete NonAdminBackup when there is no error while removing VeleroBackup and when the VeleroBackup was not in the cluster. Updated manual tests in the description of this PR that were performed.

docs/design/nab_status_update.md
mateusoliveira43, line 1: I think you can run `git restore -s master docs/design/nab_status_update.md` to undo changes in this file

mpryc: Why?
It's actually fixing the unfinished ``` from mermaid?

internal/controller/nonadminbackup_controller.go
```diff
@@ -118,6 +126,52 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if nab.Spec.DeleteVeleroBackup != nil && *nab.Spec.DeleteVeleroBackup {
+		deleted, deleteErr := function.DeleteVeleroBackup(ctx, r.Client, rLog, nab)
+		if !deleted {
+			// The Velero Object was not found
+			// Remove NonAdminBackup
+			// There is potential that Velero Sync controller will recreate Velero Backup
+			// if one exists in the s3 storage, and this will as well cause NonAdminBackup
+			// to be recreated by the NAC sync controller, but at this stage we have current
+			// proper state reflected.
+			logger.V(1).Info(fmt.Sprintf("Velero Backup not deleted for NAB: %s", nab.Name))
```
mateusoliveira43, line 150: would be interesting to add reason here?

something like `"Velero Backup %s, referenced in for %s, was not found in %s namespace and was not deleted"`

docs/design/nab_status_update.md
mateusoliveira43: git is telling us that the file was modified (in this case, last empty line was removed, which breaks CI with editorconfig), but there is no need to update it, right?

the command would copy file from master to here, and would undo all changes done to the file

internal/controller/nonadminbackup_controller.go
```diff
@@ -118,6 +126,52 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+//
+//	ctx: Context for the request.
+//	logrLogger: Logger instance for logging messages.
+//	nab: Pointer to the NonAdminBackup object.
+func (r *NonAdminBackupReconciler) DeleteVeleroBackup(ctx context.Context, logrLogger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (exitReconcile bool, errorReconcile error) {
+	rLog := log.FromContext(ctx)
+	logger := logrLogger.WithValues("DeleteVeleroBackup", nab.Namespace)
+
+	// Delete Velero Backup if the deleteVeleroBackup was set to true
+	if nab.Spec.DeleteVeleroBackup != nil && *nab.Spec.DeleteVeleroBackup {
```
mateusoliveira43, line 141: should we check here NAB status?

for example, I create a NAB with delete: true, it will be immediately deleted, right? should we only delete finished NABs?

internal/common/function/function.go
mateusoliveira43: I checked `GenerateVeleroBackupName` function, and I think my fear and Shubham's wont happen: that function is not random, right? it always create same name given same input

But still, we already have that info in NAB, why not get info there instead of generating the info again?

@mpryc the problem you pointed only happens if:
- NAB is deleted manually (without setting `delete: true` which will result in Velero Backup being deleted)

(which we will say in docs that will not work)

and user will only get problem creating NAB with same name as previous if user is quicker then NAC sync controller, right?

I am ok with current implementation, just pointing out my view :smiley:

internal/common/function/function.go
```diff
@@ -280,6 +281,41 @@ func CheckVeleroBackupLabels(backup *velerov1api.Backup) bool {
...
+
+	if err == nil {
+		err = r.Delete(ctx, &veleroBackup)
+		if err != nil {
+			logger.V(1).Info(fmt.Sprintf("Error deleting VeleroBackup: %s", veleroBackupName))
+			return false, err
+		}
+	} else if apierrors.IsNotFound(err) {
+		return false, nil
+	}
```
mateusoliveira43, line 301: I think implementation is flawed here

we ignore case that `err` is not nil and different than `apierrors.IsNotFound`, not?

that means that Velero Backup was not even retrieved, but we say it got deleted

mateusoliveira43: a suggestion is to always check if `err != nil` when dealing with errors to avoid forgetting anything

mateusoliveira43 (changes requested):
found a little error in code flow checking `err`

internal/common/function/function.go
mpryc: I don't want to have corner case where for some reason NAB did not got proper update of the Status.

To be honest we may want also add some safety feature to only delete VeleroBackup if it has the UID or other linkage to VeleroBackup from NAB object, this imo is separate design ?

There needs to be one source of truth here and I've chosen our function, where it could also be Status, so either way. Function is less check if status exists as we have name and only check if VeleroBackup exists.

internal/common/function/function.go
mpryc: 100% agree on that, will add that case.

internal/controller/nonadminbackup_controller.go
mpryc: I don't think so, if someone want to abuse spec then it's free to remove object that has been just created. Similarly to:

mkdir /tmp/test && rmdir /tmp/test

Nobody stops anyone from doing above

internal/controller/nonadminbackup_controller.go
mpryc: so this will cause NAB to be in phase Backing Off, and all the info about the referenced velero backup is in the status of that NAB, do we want to duplicate this info within debug logs?

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: I agree with @mateusoliveira43 we should only delete VeleroBackup that are in terminal phase. Currently, I don't think we can delete an in-progress VeleroBackup

api/v1alpha1/nonadminbackup_types.go
```diff
@@ -39,6 +39,10 @@ type NonAdminBackupSpec struct {
 	// BackupSpec defines the specification for a Velero backup.
 	BackupSpec *velerov1api.BackupSpec `json:"backupSpec,omitempty"`
 
+	// DeleteVeleroBackup tells the controller to remove created Velero Backup.
+	// +optional
+	DeleteVeleroBackup *bool `json:"deleteVeleroBackup,omitempty"`
```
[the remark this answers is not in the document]
shubham-pampattiwar: yeah `DeleteBackup` soudns good !

internal/common/function/function.go
shubham-pampattiwar: The NAB will not have a status only when the NAB controller is not running right ? or if the NAB controller is running and the status update call fails ? I think this makes it even more prudent that we check for the status of NAB first:
- check if NAB has status
- lookup of VeleroBackup object from NAB.status.VeleroBackupName
- VeleroBackup is in terminal state
- then delete the VeleroBackup object and then subsequently NAB
Picking up the VeleroBackupName from NAB status at the very least makes sure that VeleroBackup exists for that NAB object and Delete Call makes sense from user perspective.

mpryc: There is a time period when the NAB was created, but Backup was not yet created. If we are in that time window there is a problem with getting proper velero backup name from status. If that happens we won't reconcile again and that object will be out of sync. That is the only additional use case which I can think of when it comes to inconsistency.

shubham-pampattiwar: See thats the thing, we are not allowing Delete for such cases, grabbing the name from status would not allow a delete for such cases(because there is no VeleroBackup object in terminal phase)

mpryc: Ok, more then half of myself tend to agree with you now :-) Thanks.

shubham-pampattiwar:
Closing this in favor of https://github.com/migtools/oadp-non-admin/pull/108

PR #69, opened 2024-05-17
fix: OADP version release with NAC

mateusoliveira43:
## Why the changes were made

Update first OADP version that will contain NAC feature.

Check if new information is correct.


shubham-pampattiwar (approved):

shubham-pampattiwar (commented):
/lgtm

PR #23, opened 2024-04-03
Design documentation for the NonAdminBackup Status and Conditions

mpryc:
Two proposals for updating and keeping in sync NonAdminBackup Status and Condition fields.

The Status also consists of Velero Backup Name within NAB object.

PR #70, opened 2024-05-17
fix: Update status design with NAR

mateusoliveira43:
## Why the changes were made

Follow up of https://github.com/migtools/oadp-non-admin/pull/23 to:
- update workflow, that was changed by previous PRs
- add NonAdminRestore to design, that will be added by https://github.com/migtools/oadp-non-admin/pull/42

[Read design](https://github.com/mateusoliveira43/oadp-non-admin/blob/fix/update-status-design/docs/design/nab_and_nar_status_update.md) and see if it is clear and the code follows it.


config/crd/bases/nac.oadp.openshift.io_nonadminbackups.yaml
```diff
@@ -738,6 +738,150 @@ spec:
                       file in object storage.
                     type: integer
                 type: object
+              veleroObject:
```
mateusoliveira43, line 741: Reading design I though about this: creating a struct that contains all velero things
- name
- namespace
- status

called veleroObject but it could be veleroBackup and veleroRestore for NAB and NAR respectively

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,134 @@
...
+| Accepted | The NonAdminBackup/NonAdminRestore object was accepted by the reconcile loop, but the Velero Backup/Restore may have not yet been created |
+| Queued | The Velero Backup/Restore was created successfully. At this stage errors may still occur either from the Velero not accepting object or during backup/restore procedure. |
+
+### VeleroStatus
+
+The `VeleroStatus` field is a copy of the Velero `Backup.Status` for NonAdminBackup; and a copy of Velero `Restore.Status` for NonAdminRestore.
+
+### Velero object reference
+
+TODO
```
mateusoliveira43, line 61: TODO

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,134 @@
...
+velero backup describe -n openshift-adp nab-nacproject-c3499c2729730a
+```
+
+## Example
+
+Sample status field of a NonAdminBackup object.
+It is similar for a NonAdminRestore.
+Object passed validation and allowed to create Velero `Backup` object, but there was an error while performing backup itself:
+
+TODO
```
mateusoliveira43, line 87: TODO

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,134 @@
...
+
+This document outlines the design around updating NonAdminBackup (NAB) and NonAdminRestore (NAR) objects' Status.
+
+## NonAdminBackup and NonAdminRestore Status
+
+The `status` field of NAB and NAR objects contains the following fields:
+- `phase`
+- `conditions`
+- `veleroStatus`
+- `veleroObject` TODO
```
mateusoliveira43, line 13: TODO

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,193 @@
+# Developer Workflow: NonAdminBackup and NonAdminRestore Status Update
```
mateusoliveira43, line 1: Controller, developer seems like the developer does the work

config/crd/bases/nac.oadp.openshift.io_nonadminbackups.yaml
mateusoliveira43: applied change to the code. See if you like. If not, I can undo the change

internal/controller/nonadminbackup_controller.go
```diff
@@ -348,8 +354,8 @@ func updateNonAdminBackupVeleroBackupReference(status *nacv1alpha1.NonAdminBacku
 // updateNonAdminBackupVeleroBackupStatus sets the VeleroBackup status field in NonAdminBackup object status and returns true
 // if the VeleroBackup fields are changed by this call.
 func updateNonAdminBackupVeleroBackupStatus(status *nacv1alpha1.NonAdminBackupStatus, veleroBackup *velerov1.Backup) bool {
```
shubham-pampattiwar, line 356: Check both the status to be not nil ? (Most likely need a nil check for veleroBackup.status )

api/v1alpha1/nonadminbackup_types.go
```diff
@@ -45,20 +45,25 @@ type NonAdminBackupSpec struct {
 	LogLevel string `json:"logLevel,omitempty"`
 }
 
-// NonAdminBackupStatus defines the observed state of NonAdminBackup
-type NonAdminBackupStatus struct {
-	// VeleroBackupName references the VeleroBackup object by it's name.
+// VeleroBackup contains information of the related Velero backup object.
```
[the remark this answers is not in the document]
mateusoliveira43: OADP PR to unblock this one https://github.com/openshift/oadp-operator/pull/1546

shubham-pampattiwar (approved):

shubham-pampattiwar:
@mateusoliveira43 Whenever we do follow up commits with incremental changes, lets update the commit message to reflect those.

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,192 @@
...
+## Overview
+
+This document outlines the design around updating NonAdminBackup (NAB) and NonAdminRestore (NAR) objects' Statuses.
+
+## NonAdminBackup and NonAdminRestore Status
+
+The `status` field of NAB and NAR objects contains the following fields:
+- `phase`
+- `conditions`
+- `veleroBackup` for NAB and `veleroRestore` for NAR, which contains name, namespace and status of the related Velero object.
```
mpryc, line 12: Just to be on a very clear path. Should we have as well:
"which contains name, namespace, status and generated non-admin UUID of the corresponding Velero object."

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,192 @@
...
+
+It is always a one well defined value, that is intended to be a comprehensive state of a NAB or NAR object.
+
+Those are are the possible values for phase:
+
+| **Value** | **Description** |
+|-----------|-----------------|
+| New | *NonAdminBackup/NonAdminRestore* resource was accepted by the NAB/NAR Controller, but it has not yet been validated by the NAB/NAR Controller |
+| BackingOff | *NonAdminBackup/NonAdminRestore* resource was invalidated by the NAB/NAR Controller, due to invalid Spec. NAB/NAR Controller will not reconcile the object further, until user updates it |
+| Created | *NonAdminBackup/NonAdminRestore* resource was validated by the NAB/NAR Controller and Velero *Backup/restore* was created. The Phase will not have additional information about the *Backup/Restore* run |
```
mpryc, line 30: There was discussion (don't remember with whom) if we should add 4th and 5th Phase: `Completed` when the Velero operation completes with success and `Failed` when the backup failed. I know it's a bit of duplication with Condition, but the phase is more important for the user.

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,192 @@
...
+| lastTransitionTime | Timestamp for when this condition was created/updated. |
+| reason | Machine-readable, UpperCamelCase text indicating details about the last status transition. |
+| message | Human-readable message indicating details about the last status transition. |
+
+Those are are the possible values for `NonAdminCondition`:
+
+| **Value** | **Description** |
+|-----------|-----------------|
+| Accepted | The NonAdminBackup/NonAdminRestore object was accepted by the controller, but the Velero Backup/Restore may have not yet been created |
+| Queued | The Velero Backup/Restore was created successfully. At this stage errors may still occur either from the Velero not accepting object or during backup/restore procedure. |
```
mpryc, line 53: How about in progress/completed/failed, should we also include those?

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Velero object reference
+
+NonAdminBackup/NonAdminRestore `status` contains reference to the related Velero Backup/Restore.
+
+NonAdminBackup `status.veleroBackup` contains `name`, `namespace` and `status`.
+- `status.veleroBackup.name` represents the name of the `VeleroBackup` object.
```
mpryc, line 60: in the NAB we in my opinion we should store our generated UUID of a Velero Object. The same UUID will be stored on the Velero side but as a Label, so the user won't be able to change the status on the NAB/NAR object easily.

docs/design/nab_and_nar_status_update.md
```diff
@@ -0,0 +1,192 @@
...
+    namespace: openshift-adp
+    status:
+     ...
+```
+
+```shell
+velero backup describe -n openshift-adp nab-nacproject-c3499c2729730a
+```
+
+Similarly, NonAdminRestore `status.veleroRestore` contains `name`, `namespace` and `status`.
```
mpryc, line 79: I don't see here how the NAR is related to NAB.

docs/design/nab_and_nar_status_update.md
mateusoliveira43: I think we will, in the end, add more conditions. But when we do, we update. Otherwise I am trying to predict the future

docs/design/nab_and_nar_status_update.md
mateusoliveira43: here, we are just telling that both NAB and NAR have Velero reference in their status, not related. Is it not worded correctly?

docs/design/nab_and_nar_status_update.md
mpryc: We need to add it somewhere. Either as label or in Status. Adding in label as we discussed will require webhook prevent users to modify the label to be set/changed from the user. Adding this as status will make our life easier by not requiring webhook to prevent that, because it's the controller only who will be setting that UUID as part of Status part of the object.

docs/design/nab_and_nar_status_update.md
mpryc: Completed and Failed are the two conditions straight from the Velero `Backup` workflow, so it's not prediction but based on actual `Backup` current behavior.

docs/design/nab_and_nar_status_update.md
mpryc: Sorry for not being clear:
 1. The user creates `NonAdminBackup` let'c call it `a-uuid`, our controller creates Velero `Backup`. Both objects either in Label and/or Status, contains references with the value `a-uuid`. At this time we don't really care much about Velero `Backup` name. It should and probably will be `a-uuid`, but it's not 100% guaranteed.

2. The user would like to restore Backup. The pre-requisite is that `NonAdminBackup` object exists in the user's namespace. The existence of the `NonAdminBackup` object is guaranteed by either simply being there from the time where user created it or from the time Sync Controller created it for the user. In this situation `NonAdminRestore` object should reference `NOT` the Velero `Backup` but instead the `NonAdminBackup`. That is why I think it's incorrect to tell that Similarly it's the same as NonAdminBackup.

To rephrase NonAdminBackup should reference Velero Backup and NonAdminRestore should reference NonAdminBackup

PR #72, opened 2024-06-27
fix: use Go 1.22

mateusoliveira43:
## Why the changes were made

OADP updated Go version to `1.22`. NAC should use same Go version for compatibility.

## How the changes were made

Changed places found by `grep -Iinr '21' .`

Run project commands and check that everything works as expected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
       matrix:
         go-version: [
           "1.22",
-          "1.23",
         ]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,8 @@ jobs:
     strategy:
       matrix:
         go-version: [
-          "1.20",
-          "1.21",
           "1.22",
+          "1.23",
         ]
 
     steps:
@@ -36,7 +35,7 @@ jobs:
         run: make simulation-test
 
       - name: Go linters
-        if: ${{ matrix.go-version == '1.21' }}
+        if: ${{ matrix.go-version == '1.22' }}
         uses: golangci/golangci-lint-action@v6
         with:
           version: v1.56.2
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -18,8 +18,6 @@ jobs:
     strategy:
       matrix:
         go-version: [
-          "1.20",
-          "1.21",
           "1.22",
```
mateusoliveira43, line 21: Tried to add 1.23 as well, but got error
```
Setup go version spec 1.23
Attempting to download 1.23...
matching 1.23...
Not found in manifest.  Falling back to download directly from Go
```

I think idea here was "lets test in OADP version (1.20) for compatibility, NAC version (1.21) for sanity and next (1.22) to see if we can update when needed"

do we want still test if updates can be done or just testing on current go version is enough?

mpryc: I think that current go version is sufficient at this point. The app is containerized, so this should not be a problem?

mateusoliveira43: yeah

Velero and OADP also tests only current Go version, so I believe is ok to only test on Go version we are using (only enforcing should be that Velero, OADP and NAC use same Go version)

shubham-pampattiwar (approved):

.github/workflows/ci.yml
```diff
@@ -15,19 +15,12 @@ on:
...
-          "1.22",
-        ]
-
     steps:
       - uses: actions/checkout@v4
+
       - uses: actions/setup-go@v5
         with:
-          go-version: ${{ matrix.go-version }}
+          go-version-file: 'go.mod'
```
kaovilai, line 23: Does this repo go.mod sync with velero fork?

mateusoliveira43: currently not (it is 1.22.0 in default branch, but both will be minor synced after this PR merges)

and after an update PR merges upstream, it will be also 1.22

PR #88, opened 2024-09-25
Fix for #87 - Don't update NAB Spec from Velero Backup Spec

### LINKED ISSUE #87: Do not update the NonAdminBackup Spec from Velero Backup Spec
Do not update NonAdminBackup Spec from the VeleroSpec by reconcile loop.

This is part of #77, however #77 is about webhook to prevent the NAB being created and this issue is about Spec not being updated automatically from the Velero Backup Spec

mpryc:
Change to remove update of the NAB Spec from the Velero Backup Spec during reconcile loop.

Small nit change to add velerov1 api to import.

## Why the changes were made

Fix #87

- Running `make lint` and `make  test`
- Creating NonAdminBackup object and ensuring only Status will get updated from the original Velero Backup object. The resulting NonAdminBackup object correctly was not updated with the VeleroBackupSpec at the same time Status was updated from the Velero Backup object.

### DIFF
--- a/internal/controller/nonadminbackup_controller.go
+++ b/internal/controller/nonadminbackup_controller.go
@@ -98,7 +98,7 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
-	reconcileExit, reconcileRequeue, reconcileErr = r.UpdateSpecStatus(ctx, logger, &nab)
+	reconcileExit, reconcileRequeue, reconcileErr = r.SyncVeleroBackupWithNonAdminBackup(ctx, logger, &nab)
 	if reconcileRequeue {
 		return ctrl.Result{Requeue: true}, reconcileErr
 	} else if reconcileExit && reconcileErr != nil {
@@ -213,19 +213,17 @@ func (r *NonAdminBackupReconciler) ValidateSpec(ctx context.Context, logrLogger 
 	return false, false, nil
 }
 
-// UpdateSpecStatus updates the Spec and Status from the NonAdminBackup.
+// SyncVeleroBackupWithNonAdminBackup ensures the VeleroBackup associated with the given NonAdminBackup resource
+// is created, updated, and its status is reconciled. If the VeleroBackup does not exist, it creates a new one.
+// The function also updates the status and conditions of the NonAdminBackup resource to reflect the state
+// of the VeleroBackup.
 //
 // Parameters:
 //
 //	ctx: Context for the request.
-//	log: Logger instance for logging messages.
+//	logrLogger: Logger instance for logging messages.
 //	nab: Pointer to the NonAdminBackup object.
-//
-// The function generates the name for the Velero Backup object based on the provided namespace and name.
-// It then checks if a Velero Backup object with that name already exists. If it does not exist, it creates a new one
-// and updates NonAdminBackup Status. Otherwise, updates NonAdminBackup VeleroBackup Spec and Status based on Velero Backup object Spec and Status.
-// The function returns boolean values indicating whether the reconciliation loop should exit or requeue
-func (r *NonAdminBackupReconciler) UpdateSpecStatus(ctx context.Context, logrLogger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (exitReconcile bool, requeueReconcile bool, errorReconcile error) {
+func (r *NonAdminBackupReconciler) SyncVeleroBackupWithNonAdminBackup(ctx context.Context, logrLogger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (exitReconcile bool, requeueReconcile bool, errorReconcile error) {
 	logger := logrLogger
 
 	veleroBackupName := function.GenerateVeleroBackupName(nab.Namespace, nab.Name)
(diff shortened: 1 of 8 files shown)


internal/controller/nonadminbackup_controller_test.go
```diff
@@ -25,7 +25,7 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
-	"github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
mateusoliveira43, line 28: I like this change, but lets enforce it

In `.golangci.yml` please do the following change
```diff
@@ -47,6 +47,11 @@ linters-settings:
       limitations under the License.
   govet:
     enable-all: true
+  importas:
+    no-unaliased: true
+    alias:
+      - pkg: github.com/vmware-tanzu/velero/pkg/apis/velero/v1
+        alias: velerov1
   misspell:
     locale: US
   nakedret:
@@ -95,6 +100,7 @@ linters:
     - gosec
     - gosimple
     - govet
+    - importas
     - ineffassign
     - misspell
     - nakedret
```

This will enforce that every place that imports velero v1 package in the project, uses the `velerov1` alias

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -583,15 +533,11 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminBackup Controller",
 				ginkgo.By("Validating NonAdminBackup Spec")
 				gomega.Expect(reflect.DeepEqual(
 					nonAdminBackup.Spec.BackupSpec,
-					&v1.BackupSpec{
```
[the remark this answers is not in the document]
mpryc: I actually removed the IncludedNamespaces so we ensure that this is empty BackupSpec{}. This test isn't useless imo.

internal/controller/nonadminbackup_controller.go
```diff
@@ -308,33 +306,22 @@ func (r *NonAdminBackupReconciler) UpdateSpecStatus(ctx context.Context, logrLog
...
 	updated = updateNonAdminBackupVeleroBackupStatus(&nab.Status, &veleroBackup)
 	if updated {
 		if err := r.Status().Update(ctx, nab); err != nil {
-			veleroBackupLogger.Error(err, "NonAdminBackup BackupStatus - Failed to update")
+			veleroBackupLogger.Error(err, "Failed to update NonAdminBackup Status after VeleroBackup reconciliation")
 			return true, false, err
 		}
 
-		logger.V(1).Info("NonAdminBackup - Requeue after Status Update")
+		logger.V(1).Info("NonAdminBackup Status updated successfully, requeuing")
```
mateusoliveira43, line 320: why requeue here? The next reconcile will do nothing

Example: requeue https://github.com/migtools/oadp-non-admin/actions/runs/11031196102/job/30637414023?pr=88#step:5:110 and next reconcile does not do anything with NAB

mateusoliveira43 (changes requested):
pointed some points that need changes. Otherwise, it is good to merge already

internal/controller/nonadminbackup_controller_test.go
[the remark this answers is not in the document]
mateusoliveira43: Ok, you want to test that the spec prior to reconcile and after reconcile is the same. This does not do this. If we add a different NAB spec in the future, this breaks

What you want to do is save the spec prior to reconcile and compare with the spec after reconcile, then that is a good test!

Can you make this change then? (I would also change message in `ginkgo.By` to **Checking if NonAdminBackup Spec was not changed**)

internal/controller/nonadminbackup_controller.go
mpryc: If we agree that status update is final step in our object then we don't need to requeue, but if there is a chance there will be some extra work to be done on that updated object then we should requeue. In this case we updated NAB Status based on Velero object so I think it's safe to not re-queue ?

mpryc: If you agree with the above I will change to:

```
	veleroBackupLogger.Info("VeleroBackup already exists, verifying if NonAdminBackup Status requires update")
	updated = updateNonAdminBackupVeleroBackupStatus(&nab.Status, &veleroBackup)
	if updated {
		if err := r.Status().Update(ctx, nab); err != nil {
			veleroBackupLogger.Error(err, "Failed to update NonAdminBackup Status after VeleroBackup reconciliation")
			return true, false, err
		}

		logger.V(1).Info("NonAdminBackup Status updated successfully")
	}

	return false, false, nil
```

mateusoliveira43: I agree :)

shubham-pampattiwar (approved):

internal/controller/nonadminbackup_controller_test.go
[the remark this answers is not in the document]
mateusoliveira43: One last change: `nonAdminBackup.Spec.BackupSpec` -> `nonAdminBackup.Spec` and `scenario.spec.BackupSpec` -> `scenario.spec`

This way we ensure whole spec was not changed, and not just a field in spec

internal/controller/nonadminbackup_controller.go
```diff
@@ -213,27 +213,25 @@ func (r *NonAdminBackupReconciler) ValidateSpec(ctx context.Context, logrLogger
 	return false, false, nil
 }
 
-// UpdateSpecStatus updates the Spec and Status from the NonAdminBackup.
+// SyncVeleroBackupWithNonAdminBackup ensures the VeleroBackup associated with the given NonAdminBackup resource
+// is created, updated, and its status is reconciled. If the VeleroBackup does not exist, it creates a new one.
```
mateusoliveira43, line 217: This sentence is wrong, no? The function only creates VeleroBackup, update and status is reconcile is Velero responsibility

Would change

> is created, updated, and its status is reconciled. If the VeleroBackup does not exist, it creates a new one.

to 

> is created, if it does not exist.

mpryc: fixed.

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -421,59 +421,9 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
 			},
 			result: reconcile.Result{Requeue: true},
 		}),
-		ginkgo.Entry("When triggered by Requeue(NonAdminBackup phase created; Conditions Accepted True, Queued True; VeleroBackupStatus), should update NonAdminBackup spec BackupSpec and Requeue", nonAdminBackupSingleReconcileScenario{
```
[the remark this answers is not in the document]
mpryc: Removed.

mateusoliveira43 (approved):
/lgtm

PR #96, opened 2024-10-02
fix: Change project domain

mateusoliveira43:
## Why the changes were made

Change project domain for easier webhook integration with OADP.

Change were made following kubebuilder upgrade documentation. 

Blocked by https://github.com/openshift/oadp-operator/pull/1589

Check all places where changed (`grep -Inr "nac.oadp.openshift.io" .`)

Run the project following development documentation and check if nothing broke with the changes.


domain-change.diff
mateusoliveira43, line 1: will remove prior to merge, just to show the diff file used

shubham-pampattiwar (commented):
Please rebase this PR and also create a corresponding OADP operator PR simultaneously.

shawn-hurley:
Why are we moving the group out of nac to just oadp?

mateusoliveira43:
@shawn-hurley , @mpryc first implementations with webhook in OADP side suggested would be easier if NAC objects have same group as OADP objects

@shubham-pampattiwar waiting to see if we will need this for that implementation. If not, I will close this one without merging

shubham-pampattiwar:
yeah we need this. Basically:
- OADP Operator is responsible for installing the NAC CRDs, using `oadp` as a group makes it clear that the CRD/resource is part of the OADP ecosystem.
- Another thing would be that `oadp` as a group fir NAC CRDs keeps the CRDs logically consistent and simplifies grouping for RBAC policies, docs and user understanding.

domain-change.diff
mpryc: still in the latest force push, need to remove that.

mpryc (approved):

PR #97, opened 2024-10-03
fix: Reconcile duplication

### LINKED ISSUE #60: Split fetching Velero Backup spec and NAB spec Validation tasks

### LINKED ISSUE #62: Consider simplifying logic in reconcile loop for NAB.
Alright, what about keeping as it is, but simplify `if`s :question: 
```go
	if reconcileExit || reconcileRequeue || reconcileErr != nil{
		return ctrl.Result{Requeue: reconcileRequeue, RequeueAfter: requeueTimeSeconds * time.Second}, reconcileErr
	}
```

_Originally posted by @mateusoliveira43 in https://github.com/migtools/oadp-non-admin/pull/54#discussion_r1591400725_

mateusoliveira43:
## Why the changes were made

Fixes #62 
Fixes #60 

Run tests and check logs (`make simulation-test`) or test in a cluster following [development documentation](https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#install-from-source).

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -24,7 +24,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
-	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


internal/common/function/function.go
```diff
@@ -34,44 +34,21 @@ import (
...
-		_, _ = fmt.Println("Error merging labels:", err)
-		// TODO break?
-		return defaultLabels
-	}
-	return mergedLabels
 }
 
-// AddNonAdminBackupAnnotations return a map with both the object annotations and with the default NonAdminBackup annotations.
-// If error occurs, a map with only the default NonAdminBackup annotations is returned
-func AddNonAdminBackupAnnotations(ownerNamespace string, ownerName string, ownerUUID string, existingAnnotations map[string]string) map[string]string {
```
mpryc, line 59: @mateusoliveira43 why to remove `mergeMaps` from this function? Similarly from AddNonAdminLabels? The idea was to be able to add labels/annotations to whatever object had in a first place and propagate this to the Velero object.

Current function is completely modified, so if the user creates objects containing already existing annotations or labels which is perfectly valid scenario, those will get overridden/lost.

mpryc: The function name was `AddNonAdminBackupAnnotations` so Add NonAdminBackup annotations to the object annotations.

You are just modyfing this function to:

`GetNonAdminBackupDefaultAnnotations`

mateusoliveira43: Taking `AddNonAdminBackupAnnotations` as example

It is only used here in the code, right?
https://github.com/migtools/oadp-non-admin/blob/f99b25cc3b05ab28993294947bcf5d482bc7ee5c/internal/controller/nonadminbackup_controller.go#L258

In that place, the controller is creating the Velero Backup, so, there will never be a annotation conflict, right?

Same idea for labels.

So, I removed the check from function because they were unnecessary

mpryc: So in this case I think we don't need that function at all and we can simply specify those labels around the above code, this way we won't expose confusing public function (confising, because we don't have now confusing name that it tells it's adding labels/annotations) ?

mpryc: or rename those functions to something more meaningful - current function of it.

mateusoliveira43: yes, let's rename then

what is your suggestion?

mpryc: `GetNonAdminBackupAnnotations`

internal/common/function/function.go
```diff
@@ -34,44 +34,21 @@ import (
 	"github.com/migtools/oadp-non-admin/internal/common/constant"
 )
 
-const requiredAnnotationError = "backup does not have the required annotation '%s'"
-
-// AddNonAdminLabels return a map with both the object labels and with the default Non Admin labels.
-// If error occurs, a map with only the default Non Admin labels is returned
-func AddNonAdminLabels(labels map[string]string) map[string]string {
```
mpryc, line 41: `GetNonAdminLabels`

mpryc (approved):

Issue #90, opened 2024-09-25
`GenerateVeleroBackupName` result is not unique


mateusoliveira43:
Given different inputs, `GenerateVeleroBackupName` can return the same result. 
https://github.com/migtools/oadp-non-admin/blob/282620c4f6055e696d5d81df1af124bfd5481b6c/internal/common/function/function.go#L111

This can result in bad user experience, as users can end using other users backup (or when UUID solution is implemented, https://github.com/migtools/oadp-non-admin/issues/89, prevent users from using NonAdminBackup with certain names)

From what I studied, given 2 namespaces that have long names (more than 239 chars), if their names differ only in position 240 or above, any NonAdminBackup that have the same same in these namespaces, the associated VeleroBackup name will be the same.

Example

this
> namespace: mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm1
> NonAdminBackup: example

and this input
> namespace: mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm2
> NonAdminBackup: example

return the same result

## Further investigation

Need to confirm no other scenarios will generate equal names.

kaovilai:
might help to look at https://github.com/kubernetes/enhancements/blob/61b21e5105e5fda4d6c53821ff82dd39a1a98f07/keps/sig-api-machinery/4420-retry-generate-name/README.md

mpryc:
[...]

PR #98, opened 2024-10-03
Related 89: Function to generate custom UUID for NAC objects

### LINKED ISSUE #89: Use custom UUID for NAB Object to Velero Backup Object mapping instead of the Kube UUID
We use Kube generated UUID for establishing mapping between NAB Object and Velero Backup Object. Instead of this, use a custom UUID for establishing mapping. This will help avoid UUID collisions.

mpryc:
This function will be used to link between NAB and Backup and possibly other parts of the objects within Non-Admin-Backup.

## Why the changes were made

Function that will be used to fix #89 

Run unit tests and ensure many corner cases are handled.

### DIFF
--- a/internal/common/constant/constant.go
+++ b/internal/common/constant/constant.go
@@ -39,6 +39,9 @@ const (
 // EmptyString defines a constant for the empty string
 const EmptyString = ""
 
+// BackupNameDelimiter defines character that is used to separate name aprts
+const BackupNameDelimiter = "-"
+
 // TrueString defines a constant for the True string
 const TrueString = "True"
 
--- a/internal/common/constant/constant.go
+++ b/internal/common/constant/constant.go
@@ -39,8 +39,8 @@ const (
 // EmptyString defines a constant for the empty string
 const EmptyString = ""
 
-// BackupNameDelimiter defines character that is used to separate name aprts
-const BackupNameDelimiter = "-"
+// NameDelimiter defines character that is used to separate name parts
+const NameDelimiter = "-"
 
 // TrueString defines a constant for the True string
 const TrueString = "True"
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/google/uuid"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
@@ -134,6 +135,51 @@ func GenerateVeleroBackupName(namespace, nabName string) string {
 			namespace = namespace[:maxNamespaceLength]
 		}
 		veleroBackupName = fmt.Sprintf("%s-%s-%s", constant.VeleroBackupNamePrefix, namespace, nameHash)
+	}
+
+	return veleroBackupName
+}
+
+// GenerateVeleroBackupNameWithUUID generates a Velero backup name based on the provided namespace and NonAdminBackup name.
+// It includes a UUID suffix. If the name exceeds the maximum length, it truncates nabName first, then namespace.
+func GenerateVeleroBackupNameWithUUID(namespace, nabName string) string {
+	// Generate UUID suffix
+	uuidSuffix := uuid.New().String()
+
+	// Build the initial backup name based on the presence of namespace and nabName
+	veleroBackupName := uuidSuffix
+	if len(nabName) > 0 {
+		veleroBackupName = nabName + constant.BackupNameDelimiter + veleroBackupName
+	}
+	if len(namespace) > 0 {
+		veleroBackupName = namespace + constant.BackupNameDelimiter + veleroBackupName
+	}
+
+	// Ensure the name is within the character limit
+	maxLength := validation.DNS1123SubdomainMaxLength
+
+	if len(veleroBackupName) > maxLength {
+		// Calculate remaining length after UUID
+		remainingLength := maxLength - len(uuidSuffix)
+
+		delimeterLength := len(constant.BackupNameDelimiter)
+
+		// Subtract two delimiter lengths to avoid a corner case where the namespace
+		// and delimiters leave no space for any part of nabName
+		if len(namespace) > remainingLength-delimeterLength-delimeterLength {
+			namespace = namespace[:remainingLength-delimeterLength-delimeterLength]
+			veleroBackupName = namespace + constant.BackupNameDelimiter + uuidSuffix
+		} else {
+			remainingLength = remainingLength - len(namespace) - delimeterLength - delimeterLength
+			nabName = nabName[:remainingLength]
+			veleroBackupName = uuidSuffix
+			if len(nabName) > 0 {
+				veleroBackupName = nabName + constant.BackupNameDelimiter + veleroBackupName
+			}
+			if len(namespace) > 0 {
+				veleroBackupName = namespace + constant.BackupNameDelimiter + veleroBackupName
+			}
+		}
 	}
 
 	return veleroBackupName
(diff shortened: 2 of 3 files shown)


mpryc:
@mateusoliveira43 note this function is separate to currently used one so we can safely move to this new one without breaking working current controller code.

internal/common/function/function.go
```diff
@@ -139,6 +140,51 @@ func GenerateVeleroBackupName(namespace, nabName string) string {
 	return veleroBackupName
 }
 
+// GenerateVeleroBackupNameWithUUID generates a Velero backup name based on the provided namespace and NonAdminBackup name.
+// It includes a UUID suffix. If the name exceeds the maximum length, it truncates nabName first, then namespace.
+func GenerateVeleroBackupNameWithUUID(namespace, nabName string) string {
```
kaovilai, line 145: I assume whatever calls this func is going to call it again if on create it found a duplicate velero backup name?

internal/common/constant/constant.go
```diff
@@ -39,6 +39,9 @@ const (
 // EmptyString defines a constant for the empty string
 const EmptyString = ""
 
+// BackupNameDelimiter defines character that is used to separate name aprts
+const BackupNameDelimiter = "-"
```
shubham-pampattiwar, line 43: NIT: We may re-use this later e.g: restore etc
Rename it to `NameDelimiter`

internal/common/function/function.go
```diff
@@ -139,6 +140,51 @@ func GenerateVeleroBackupName(namespace, nabName string) string {
...
+	// Generate UUID suffix
+	uuidSuffix := uuid.New().String()
+
+	// Build the initial backup name based on the presence of namespace and nabName
+	veleroBackupName := uuidSuffix
+	if len(nabName) > 0 {
+		veleroBackupName = nabName + constant.BackupNameDelimiter + veleroBackupName
+	}
+	if len(namespace) > 0 {
```
shubham-pampattiwar, line 154: Maybe we can drop using namespace in VeleroBackupName. It will anyways be part of metadata/spec ?

mpryc: I think it's useful to quickly filter the namespace from which Backup was created. So if we have multiple users with admin rights that have separate namesapces it allows to quickly sort/identify which Backups are from which ns. Of course this will be also available as additional info via annotation or label, but here is just another step to not have one list of unreadable by human backup names (uuid's only as example).

internal/common/function/function.go
mpryc: No. This function is going to be called once and the generated UUID will be unique. This UUID will get stored in both NAB object and other objects associated with that NAB such as Backup.

kaovilai: How can you be sure?
Didn't we say we wanna retry on conflict?

Either way that should be done on create call failure due to already existing backup of the same name.

kaovilai: > [simply use our generated UUID as the Backup object name with retry generate](https://github.com/migtools/oadp-non-admin/issues/90#issuecomment-2376188425)

> [Yeah, a retry on velero backup name seems a good approach](https://github.com/migtools/oadp-non-admin/issues/90#issuecomment-2376831858)

mpryc: @kaovilai that is exactly how this should be used:

1. Whoever calls this function implements retry logic
2. The NonAdminBackup calls this function to get the Backup name as UUID - to make it easier let's call it UUID_1 if it fails due to unlikely event of collision it does retry.
3. Once 2 succeed the Backup is generated with the name that corresponds to this name. Moreover the label is added with exactly same name (UUID_1).
4. The UUID_1 is also stored in NonAdminBackup object as label value.
5. When next reconcile comes to NonAdminBackup it checks if there is label with some UUID. We get in our case UUID_1 so we can get the Velero Backup object with UUID_1. It works vice-versa, whenever Backup object requires to update NAB object it can be identified by this UUID_1.

kaovilai: Sounds good.

internal/common/function/function.go
mateusoliveira43: If we are using a constant prefix in NAB names, this function needs to change

way it is, we call it and then check if result is valid; if not, we will retry

but on retry, only thing that will change is name suffix, which has the UUID

Correct way should be to create a function that generates the prefix (in this case `<nab-name>-<nab-namespace>-`, truncated) if needed

then, inside retry loop, we create UUID and add result of previous function and try to create velero backup. If it fails, we just retry with new UUID, prefix was already created and will not change

mateusoliveira43:
@mpryc I do not like this approach, we would be merging unused code

We already have integration tests in place, when adding new way of generating NAB name, we should replace old way in same PR (and tests can tell us if new implementation works as expected)

internal/common/function/function.go
mpryc: It's essentially the same, just a different approach to writing the function. I prefer to encapsulate everything in a single function to generate the full name. If we go with your approach, we'd need to:

1. Create function to get `<nab-name>-<nab-namespace>-` with additional parameter of max length (we do not know if UUID will always have static length)
2. Make logic to merge above with generated UUID
3. Make logic to retry if the name exists inside reconcile

With current implementation we skip entirely point 2 making it easier within reconcile.

mateusoliveira43: from this https://pkg.go.dev/github.com/google/uuid#Parse it is always 36 chars length

mpryc: Correct if we take RFC 4122, then is always 36 but really it should not be split between two functions to create string.

internal/common/function/function.go
shubham-pampattiwar, line 145: I think we are gonna use a similar function for NorAdmin Restore Objects ,right ? Would it not make sense to modify this function to just `GenerateNameWithUUID` ? (Refactor it so that there is not context of Backup/Restore, just namespace and NacObjectName as parameters)

kaovilai: We can always follow up with a refactor where this func calls a more common func

mpryc:
Closing as I am preparing new PR with this change included.

internal/common/function/function.go
mpryc: I am closing this PR, new function is generic and will be included in other PR

mpryc:
```
// GenerateNacObjectNameWithUUID generates a unique name based on the provided namespace and object origin name.
// It includes a UUID suffix. If the name exceeds the maximum length, it truncates nacName first, then namespace.
func GenerateNacObjectNameWithUUID(namespace, nacName string) string {
	// Generate UUID suffix
	uuidSuffix := uuid.New().String()

	// Build the initial name based on the presence of namespace and nacName
	nacObjectName := uuidSuffix
	if len(nacName) > 0 {
		nacObjectName = nacName + constant.NameDelimiter + nacObjectName
	}
	if len(namespace) > 0 {
		nacObjectName = namespace + constant.NameDelimiter + nacObjectName
	}

	if len(nacObjectName) > constant.MaximumNacObjectNameLength {
		// Calculate remaining length after UUID
		remainingLength := constant.MaximumNacObjectNameLength - len(uuidSuffix)

		delimeterLength := len(constant.NameDelimiter)

		// Subtract two delimiter lengths to avoid a corner case where the namespace
		// and delimiters leave no space for any part of nabName
		if len(namespace) > remainingLength-delimeterLength-delimeterLength {
			namespace = namespace[:remainingLength-delimeterLength-delimeterLength]
			nacObjectName = namespace + constant.NameDelimiter + uuidSuffix
		} else {
			remainingLength = remainingLength - len(namespace) - delimeterLength - delimeterLength
			nacName = nacName[:remainingLength]
			nacObjectName = uuidSuffix
			if len(nacName) > 0 {
				nacObjectName = nacName + constant.NameDelimiter + nacObjectName
			}
			if len(namespace) > 0 {
				nacObjectName = namespace + constant.NameDelimiter + nacObjectName
			}
		}
	}

	return nacObjectName
}
```

PR #101, opened 2024-10-14
Fix for #89 - Use custom UUID for NAB Object

### LINKED ISSUE #89: Use custom UUID for NAB Object to Velero Backup Object mapping instead of the Kube UUID

### LINKED ISSUE #90: `GenerateVeleroBackupName` result is not unique

mpryc:
With this change a new UUID is generated to reference parent/child relationship between objects in the Non Admin Controller use cases.

The first consumer of this UUID is a Velero Backup, created when the NonAdminBackup object is reconciled.

The NonAdminBackup object generates the NAC UUID and stores it in its Status. This prevents users from modifying it. The UUID is later used to create the Velero Backup during reconciliation.

While the NAC UUID is currently used as the Velero Backup name, this is not required, as the UUID is also stored as a Velero Backup label, which is used during the reconcile loop. Usage of NAC UUID as Velero Backup name is to easy it's creation.

This PR also includes small changes to fix linting issues of the code, as well reworks the tests to properly take advantage of gingko BeforeEach function.

## Why the changes were made

Fixes #89 
Fixes #90


mpryc:
@shubham-pampattiwar FYI please review

internal/common/constant/constant.go
```diff
@@ -39,9 +42,12 @@ const (
...
+
 // TrueString defines a constant for the True string
 const TrueString = "True"
 
-// VeleroBackupNamePrefix represents the prefix for the object name generated
-// by the NonAdminController
-const VeleroBackupNamePrefix = "nab"
+// MaximumNacObjectNameLength represents Generated Non Admin Object Name and
+// must be below 63 characters, because it's used within object Label Value
+const MaximumNacObjectNameLength = validation.DNS1123LabelMaxLength
```
mateusoliveira43, line 53: why redefine const?

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
+	if !checkAnnotationOrLabelsValueIsValid(objLabels, constant.NabOriginUUIDLabel) {
 		return false
 	}
-	if !checkAnnotationValueIsValid(annotations, constant.NabOriginNameAnnotation) {
+
+	annotations := obj.GetAnnotations()
+	if !checkAnnotationOrLabelsValueIsValid(annotations, constant.NabOriginNamespaceAnnotation) {
 		return false
 	}
-	// TODO what is a valid uuid?
```
mateusoliveira43, line 122: I think this is a bad change

First, the function does check for max label length

It also does not check if UUID value is valid

I would not change `checkAnnotationValueIsValid` function logic

Correct approach here is to create new function to check if label value is valid and also check if UUID value is valid, by using Parse function https://pkg.go.dev/github.com/google/uuid@v1.6.0#Parse, right?

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
+
+	// Attempt to list objects with the specified label
+	if err := clientInstance.List(ctx, objectList, &client.ListOptions{
+		LabelSelector: labelSelector,
+		Namespace:     namespace,
+	}); err != nil {
+		return fmt.Errorf("failed to list objects in namespace '%s': %w", namespace, err)
+	}
+
+	return nil
```
mateusoliveira43, line 140: I would call this function something like `GetSingleObjectByLabel` and do validation here 

For case no items are found, I would use `NewNotFound` function https://pkg.go.dev/k8s.io/apimachinery@v0.29.0/pkg/api/errors#NewNotFound to keep behavior in controllers

internal/common/constant/constant.go
mpryc: Because this is more generic now and rename is to reflect it's current purpose..

internal/common/function/function.go
mpryc: The Non admin UUID is not part of annotation, it's part of label now. Also it's not a valid UUID as we add more to it (prefix).

internal/common/function/function.go
mpryc: No, this is not how it works. This particular function is generic one that returns all the objects within that label. Later the function `GetVeleroBackupByLabel` is consuming that generic function and have single object logic. I don't want to call it Single, because it's single by name and not the `GetVeleroBackupsByLabel` which would be plural.

mpryc: I've moved this function to be more generic for use in restore as well in the future.

mpryc: I don't want to use proposed function, because we do not have knowledge *why* the error when listing the objects was raised. It's an error from the `List()` function, so there may be multiple reasons why it failed to List and even if the objects are in the cluster the error still may happen.

internal/common/constant/constant.go
mpryc: It's used 8 times (2x in the code and 6x in tests).

internal/common/function/function.go
mpryc: RE: "First, the function does check for max label length"

It does by comparing it in `if len(nacObjectName) > constant.MaximumNacObjectNameLength {`

mateusoliveira43: 253 char limitation for annotation is needed here because the annotations values are kubernetes names, so if it longer than that, we can not create object with that value (sync controller for example)

mpryc: Can you point me to the code where currently annotation value is used as the base for *creating*  k8s object with the name as the value ? I don't think there is, we are using it to store already existing name but not to create.

api/v1alpha1/nonadminbackup_types.go
```diff
@@ -51,9 +51,9 @@ type VeleroBackup struct {
 	// +optional
 	Status *velerov1.BackupStatus `json:"status,omitempty"`
 
-	// name references the Velero backup by it's name.
+	// nabnacuuid references the Non Admin Backup object by it's nacUUID.
 	// +optional
-	Name string `json:"name,omitempty"`
+	NabNacUUID string `json:"nabnacuuid,omitempty"`
```
shubham-pampattiwar, line 56: I think we should keep the VB Name as well as the UUID.

internal/common/constant/constant.go
shubham-pampattiwar: IIUC, what @mateusoliveira43 is trying to say is why not use the `validation.DNS1123LabelMaxLength` const directly.

internal/common/function/function.go
shubham-pampattiwar: Yeah lets cascade the list err as is for transparency.

api/v1alpha1/nonadminbackup_types.go
mpryc: In our internal implementation the VB Name is same as UUID, so I am unsure if we want to duplicate that at this point of time? I am happy to add it if you think so:

https://github.com/migtools/oadp-non-admin/blob/97dacc63f1f009c9b6422b02165438614d482a29/internal/controller/nonadminbackup_controller.go#L251

Also please not my comment on:
https://github.com/migtools/oadp-non-admin/issues/90#issuecomment-2412374910

I think we could close 90 as well with this if we agree that retry is not required anymore (almost impossible collision in our case).

internal/common/constant/constant.go
mpryc: Yes, and my explanation is that this constant is used 8 times in the code (2 in the main code, 6 in tests), so it's much easier to modify it in one place if we ever need to adjust the length limit. I'm happy to use validation.DNS1123LabelMaxLength directly, but this is similar to other constants we use in the code. For example, why not just use "" instead of defining a TrueString constant, or use - directly instead of defining a delimiter? Similarly, why not use 63 directly instead of creating this constant?

The point is that defining MaximumNacObjectNameLength makes the code more manageable. We are setting a constraint specific to our application, which happens to align with an external one (the 63-character limit), but this allows us to easily modify it if needed in the future. By controlling our own constant, we avoid relying too heavily on the "upstream" constraint, which might change or have different context in the future.

Additionally, the constant comes with a comment explaining why this 63-character limit is relevant for NAC object names in our context. Without it, we'd need to repeat this explanation in every place where we use the raw constant, which could make the code harder to maintain and understand.

internal/common/function/function.go
mpryc: @shubham-pampattiwar by saying lets cascade you agree current implementation is correct or it needs changes?

I've implemented the function `GetObjectByLabel` to do just it's job, so no logic involved in the body of that function. It's currently used by:

https://github.com/migtools/oadp-non-admin/blob/97dacc63f1f009c9b6422b02165438614d482a29/internal/common/function/function.go#L150-L161

And the above `GetVeleroBackupByLabel` does cascade errors and lists accordingly.

We do not know what other parts such as sync controller or NAR controller will do with the situation where no objects based on the label value are found so it's up to the consumer function to decide how to handle such situation. I believe current implementation is generic and correct, because it separates `get me some object X` from logical function that handles the result for the X.

It would be different implementation if that function was not part of our `common` functions and specific **only** to the NAB object.

internal/common/function/function.go
```diff
@@ -77,38 +77,101 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
+}
+
+// GetVeleroBackupByLabel retrieves a VeleroBackup object based on a specified label within a given namespace.
+// It returns the VeleroBackup only when exactly one object is found, throws an error if multiple backups are found,
+// or returns nil if no matches are found.
+func GetVeleroBackupByLabel(ctx context.Context, clientInstance client.Client, namespace string, labelValue string) (*velerov1.Backup, error) {
+	veleroBackupList := &velerov1.BackupList{}
+
+	// Call the generic GetObjectByLabel function
+	if err := GetObjectByLabel(ctx, clientInstance, namespace, constant.NabOriginUUIDLabel, labelValue, veleroBackupList); err != nil {
```
shubham-pampattiwar, line 150: Lets just rename the function to List-something

internal/common/function/function.go
shubham-pampattiwar: Current implementation looks good, just update the func name to reflect the action.

api/v1alpha1/nonadminbackup_types.go
shubham-pampattiwar: Hmm I was thinking this field should reflect both the Name and UUID. Lets rename it to something like: `identifier` , `nameUUID` or `idName` or something.

shubham-pampattiwar: ok, works for me.

internal/common/function/function.go
mateusoliveira43: I agree with that (please link issue #90 to this PR), but this discussion is another problem (validation of annotation inside velero backup predicate)

shubham-pampattiwar: updated the PR description to link #90

internal/common/function/function.go
mpryc: Updated generic function name to:

`ListLabeledObjectsInNamespace`

However reading the code again I think the `GetVeleroBackupByLabel` is correct. It only returns object when exactly one is found, so it does not list velero backups by label.

mpryc:
@shubham-pampattiwar @mateusoliveira43 Added commit which I hope answers all comments.

internal/common/function/function.go
```diff
@@ -119,9 +119,9 @@ func GenerateNacObjectNameWithUUID(namespace, nacName string) string {
 	return nacObjectName
 }
 
-// GetObjectByLabel retrieves a list of Kubernetes objects of a specified type based on a label within a given namespace.
-// It returns a slice of the specified object type and an error.
-func GetObjectByLabel(ctx context.Context, clientInstance client.Client, namespace string, labelKey string, labelValue string, objectList client.ObjectList) error {
+// ListLabeledObjectsInNamespace retrieves a list of Kubernetes objects in a specified namespace
+// that match a given label key-value pair.
+func ListLabeledObjectsInNamespace(ctx context.Context, clientInstance client.Client, namespace string, labelKey string, labelValue string, objectList client.ObjectList) error {
```
shubham-pampattiwar, line 124: NIT: `ListObjectsByLabel`

shubham-pampattiwar (approved):

internal/common/function/function.go
mpryc: I wanted to be super precise what this function does ;)

mpryc: Changed to `ListObjectsByLabel`

shubham-pampattiwar (approved):

docs/design/Non_Admin_Controller_design.md
mateusoliveira43, line 1: This diagram is wrong

There is a reconcile start without reconcile exit

There are 2 wrong status update (one saying it only changes UUID, but it also changes status.conditions.veleroBackup.namespace; and other saying it changes phase, conditions and status.veleroBackup, but it only changes phase and conditions)

please also follow pattern (rectangles for status update)

api/v1alpha1/nonadminbackup_types.go
```diff
@@ -51,9 +51,9 @@ type VeleroBackup struct {
 	// +optional
 	Status *velerov1.BackupStatus `json:"status,omitempty"`
 
-	// name references the Velero backup by it's name.
+	// nameuuid references the Velero Backup object by it's label containing same NameUUID.
 	// +optional
-	Name string `json:"name,omitempty"`
+	NameUUID string `json:"nameuuid,omitempty"`
```
mateusoliveira43, line 56: I do not like this change

This is still the Velero backup name

Is useful to user to know how the name is generated?

internal/controller/nonadminbackup_controller.go
```diff
@@ -195,45 +225,57 @@ func (r *NonAdminBackupReconciler) validateSpec(ctx context.Context, logger logr
...
 				Labels:      function.GetNonAdminLabels(),
 				Annotations: function.GetNonAdminBackupAnnotations(nab.ObjectMeta),
 			},
 			Spec: *backupSpec,
 		}
 
+		// Add NonAdminBackup's veleroBackupNameUUID as the label to the VeleroBackup object
+		// We don't add this as an argument of GetNonAdminLabels(), because there may be
+		// situations where NAC object do not require NabOriginUUIDLabel
+		veleroBackup.Labels[constant.NabOriginNameUUIDLabel] = veleroBackupNameUUID
```
mateusoliveira43, line 262: this does not make sense to me, it is duplication of information (we can get this info from the name of the object, that is exactly the same)

mpryc: This is intended to improve usability for users, primarily administrators.

api/v1alpha1/nonadminbackup_types.go
shubham-pampattiwar: The field indicates that it stores the VB name (ux) as well as the information that it is also the UUID (dev logic)

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: yes labels help in managing and filtering objects.

docs/design/Non_Admin_Controller_design.md
shubham-pampattiwar: @mpryc Could please update fix this design doc change that @mateusoliveira43 is pointing to ?

shubham-pampattiwar (dismissed):

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
-		return ""
+// GenerateNacObjectNameWithUUID generates a unique name based on the provided namespace and object origin name.
+// It includes a UUID suffix. If the name exceeds the maximum length, it truncates nacName first, then namespace.
+func GenerateNacObjectNameWithUUID(namespace, nacName string) string {
+	// Generate UUID suffix
+	uuidSuffix := uuid.New().String()
+
+	// Build the initial name based on the presence of namespace and nacName
+	nacObjectName := uuidSuffix
+	if len(nacName) > 0 {
```
mateusoliveira43, line 88: these validations are overkill. Only place (not in tests file) this is called, is inside nab reconcile, after the object was get in the cluster. And a Kubernetes object can not have empty name or namespace, right?

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
-// If the resulting name exceeds the maximum Kubernetes name length, it truncates the namespace to fit within the limit.
-func GenerateVeleroBackupName(namespace, nabName string) string {
-	// Calculate a hash of the name
-	hasher := sha256.New()
-	_, err := hasher.Write([]byte(nabName))
-	if err != nil {
-		return ""
+// GenerateNacObjectNameWithUUID generates a unique name based on the provided namespace and object origin name.
+// It includes a UUID suffix. If the name exceeds the maximum length, it truncates nacName first, then namespace.
+func GenerateNacObjectNameWithUUID(namespace, nacName string) string {
```
mateusoliveira43, line 82: this name does not make sense to me. This does not generate NAB/NAR name, it generates Velero backup/restore name, right? `GenerateVeleroObjectName` makes more sense?

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
+	// Generate UUID suffix
+	uuidSuffix := uuid.New().String()
```
mateusoliveira43, line 84: from function comment

> New creates a new random UUID or panics. New is equivalent to the expression

I would handle panic or at least add a TODO to do this later

mpryc: I can see the function may panic. This is only for the UUID4 implementation: 

https://github.com/google/uuid/blob/0e97ed3b537927cb4afea366bc4cc36f6eb37e75/version4.go#L9

As well in the doc generated from the above source code:
  https://pkg.go.dev/github.com/google/uuid#NewString

The Panic may originate from `Must` function and randomness source failure:
https://github.com/google/uuid/blob/0e97ed3b537927cb4afea366bc4cc36f6eb37e75/uuid.go#L45

Which is part of "crypto/rand".

It's highly unlikely the panic will ever occur and most likely it's caused by issues with the operating system or hardware (e.g., the random device /dev/urandom is unavailable or malfunctions).

I really don't know if we should be such detailed to handle this panic inside our code (we can by implementing defer function to handle panic). What will happen in unlikely event of panic is that  reconcile loop will crash and then it will be retried by the Kubernetes controller-runtime, so I think we are very safe here without additional wrappers around that case.

internal/common/function/function.go
mateusoliveira43, line 82: I would change function implementation to make it more readable (did not use the consts you defined, but they can be used here)
```go
func Generate(namespace, name string) string {
	uuidSuffix := uuid.New().String()

	remainingLength := validation.DNS1123LabelMaxLength - len(uuidSuffix) - len("--")

	if len(namespace+name) > remainingLength {
		if len(namespace) >= remainingLength {
			return fmt.Sprintf("%s-%s", namespace[:remainingLength], uuidSuffix)
		}
		remainingLength := remainingLength - len(namespace)
		return fmt.Sprintf("%s-%s-%s", name[:remainingLength], namespace, uuidSuffix)
	}

	return fmt.Sprintf("%s-%s-%s", namespace, name, uuidSuffix)
}
```

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
-	maxNamespaceLength := validation.DNS1123SubdomainMaxLength - usedLength
-	// Ensure the name is within the character limit
-	if len(namespace) > maxNamespaceLength {
-		// Truncate the namespace if necessary
-		return fmt.Sprintf("%s-%s-%s", constant.VeleroBackupNamePrefix, namespace[:maxNamespaceLength], nameHash)
+// ListObjectsByLabel retrieves a list of Kubernetes objects in a specified namespace
+// that match a given label key-value pair.
+func ListObjectsByLabel(ctx context.Context, clientInstance client.Client, namespace string, labelKey string, labelValue string, objectList client.ObjectList) error {
+	// Validate input parameters
+	if namespace == constant.EmptyString || labelKey == constant.EmptyString || labelValue == constant.EmptyString {
```
mateusoliveira43, line 126: I also think this validation is overkill. This is not a function that user with be able to use, we call it inside `GetVeleroBackupByLabel`, which would trigger this error only if developer calls the function wrong

internal/common/function/function_test.go
```diff
@@ -115,36 +122,184 @@ func TestValidateBackupSpec(t *testing.T) {
...
+			name:      "Truncate nabName due to length",
+			namespace: "some",
+			nabName:   strings.Repeat("q", constant.MaximumNacObjectNameLength+10), // too long for DNS limit
+		},
+		{
+			name:      "Truncate very long namespace and very long name",
+			namespace: strings.Repeat("w", constant.MaximumNacObjectNameLength+10),
+			nabName:   strings.Repeat("e", constant.MaximumNacObjectNameLength+10),
+		},
+		{
```
mateusoliveira43, line 146: all the tests cases with empty name or namespace make no sense to me. As I commented in another place, these cases can not happen in real scenario of reconcile

internal/common/function/function_test.go
```diff
@@ -115,36 +122,184 @@ func TestValidateBackupSpec(t *testing.T) {
...
+
+			// Check if no double hyphens are present
+			if strings.Contains(result, "--") {
+				t.Errorf("Generated name contains double hyphens: %s", result)
+			}
+		})
+	}
+}
+
+func TestGetVeleroBackupByLabel(t *testing.T) {
```
mateusoliveira43, line 202: why not use testEnv instead of fake client?

internal/controller/nonadminbackup_controller.go
```diff
@@ -195,45 +225,57 @@ func (r *NonAdminBackupReconciler) validateSpec(ctx context.Context, logger logr
 //	ctx: Context for the request.
 //	logger: Logger instance for logging messages.
 //	nab: Pointer to the NonAdminBackup object.
-func (r *NonAdminBackupReconciler) syncVeleroBackupWithNonAdminBackup(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
-	veleroBackupName := function.GenerateVeleroBackupName(nab.Namespace, nab.Name)
-	if veleroBackupName == constant.EmptyString {
-		return false, errors.New("unable to generate Velero Backup name")
+func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	if nab.Status.VeleroBackup == nil || nab.Status.VeleroBackup.NameUUID == constant.EmptyString {
```
mateusoliveira43, line 229: I think this is also overkill. This part is only run after `setBackupUUIDInStatus`, which ensures this does not happen, right?

docs/design/Non_Admin_Controller_design.md
```diff
@@ -139,17 +139,17 @@ This design intends to enable non-admin users the ability to perform Backup and
...
+    - **NAB controller validates the NAB CR and then creates a corresponding Velero Backup CR:** When the NonAdminBackup controller detects a new or modified NonAdminBackup object, it creates or updates a corresponding Velero Backup object within the OADP Namespace using the information provided in the `backupSpec` field of the NonAdminBackup object. The resulting Backup object is named as `<namespace>-<name>-<hash>`, where the `<namespace>` is the NonAdminBackup namespace, `<name>` is the NonAdminBackup name and the `<hash>` is the generated UUID (version 4). If the resulting Backup object name is too long the name is adapted to 63 characters limit strpping first characters from name and then from namespace. The resulting Backup object is labeled and annotated with the following additional metadata:
```
mateusoliveira43, line 142: typo strpping ?

docs/design/nab_and_nar_status_update.md
```diff
@@ -124,69 +124,107 @@ It is similar for a NonAdminRestore.
...
-conditionsAcceptedTrue["`status.conditions[Accepted] to **True**`"] -. Requeue: false, err .- reconcileStart2
-conditionsAcceptedTrue ==>
+questionPhaseStatusSetToNew{"Is status.phase: **New** ?"};
+questionConditionAcceptedTrue{"Is status.conditions[Accepted]: **True** ?"};
+questionIsNabValid{"Is NonAdminBackup Spec valid ?"};
+questionPhaseStatusSetToBackingOff{"Is status.phase: **BackingOff**
+ and status.conditions[Accepted]: **False** ?"};
+questionPhaseStatusSetToCreated{"Is status.phase: **Created**
+ and status.conditions[BackupScheduled]: **True** ?"};
+questionStatusVeleroBackupUUID{"Is status.VeleroBackup.NameUUID existing and valid ?"};
```
mateusoliveira43, line 147: the reconcile loop does not validate UUID, right?

docs/design/nab_and_nar_status_update.md
```diff
@@ -124,69 +124,107 @@ It is similar for a NonAdminRestore.
...
+questionSuccess{"Success ?"};
+questionSuccessCreateVB{"Success ?"}
+questionSuccessWithTerminalError{"Success ?"};
+questionSuccessWithNoRequeue{"Success ?"};
+questionSuccessGetVB{"Error ?"};
```
mateusoliveira43, line 152: all questions are `success?` except this one. Better to be consistent?

docs/design/nab_and_nar_status_update.md
```diff
@@ -124,69 +124,107 @@ It is similar for a NonAdminRestore.
...
+requeueFalseNil[\"Requeue: false"/];
+requeueFalseErr[\"Requeue: false, error"/];
+requeueFalseTerminalErr[\"Requeue: false, Terminal error"/];
 
-reconcileStart2[/Reconcile start\] ==>
+reconcileStartAcceptedPredicate[/Reconcile start\];
 
-conditionsAcceptedTrue["`status.conditions[Accepted] to **True**`"] -. Requeue: false, err .- reconcileStart2
-conditionsAcceptedTrue ==>
+questionPhaseStatusSetToNew{"Is status.phase: **New** ?"};
```
mateusoliveira43, line 140: I would change all question that are not `success?` to use rectangle like shapes, otherwise they occupy to much space in diagram

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -465,69 +491,94 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminBackup Controller",
...
 					types.NamespacedName{
-						Name:      function.GenerateVeleroBackupName(nonAdminNamespaceName, testNonAdminBackupName),
-						Namespace: oadpNamespaceName,
+						Name:      nonAdminBackup.Status.VeleroBackup.NameUUID,
+						Namespace: oadpNamespace,
 					},
 					veleroBackup,
 				)).To(gomega.Succeed())
-				veleroBackup.Status.Phase = velerov1.BackupPhaseCompleted
-				// TODO can not call .Status().Update() for veleroBackup object: backups.velero.io "name..." not found error
```
mateusoliveira43, line 507: please do not remove comment. If you want, you update it, with the reason

> // can not call .Status().Update() for veleroBackup object https://github.com/vmware-tanzu/velero/issues/8285

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -465,69 +491,94 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminBackup Controller",
...
-				gomega.Expect(k8sClient.Update(ctx, veleroBackup)).To(gomega.Succeed())
+
+				veleroBackup.Status = velerov1.BackupStatus{
+					Phase: velerov1.BackupPhaseCompleted,
+				}
+
+				gomega.Expect(k8sClient.Update(ctxTimeout, veleroBackup)).To(gomega.Succeed())
+
+				time.Sleep(1 * time.Second)
```
mateusoliveira43, line 556: this is unnecessary, right?

docs/design/nab_and_nar_status_update.md
mpryc: This is special one, where we can get 3 states in one decision block or split it by success and error. I wanted to have this in the 2 blocks to easy distinguish between those states as they result in different action.

internal/common/function/function.go
```diff
@@ -77,66 +77,125 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup) error {
...
+		return nil, err
+	}
+
+	switch len(veleroBackupList.Items) {
+	case 0:
+		return nil, nil // No matching VeleroBackup found
+	case 1:
+		return &veleroBackupList.Items[0], nil // Found 1 matching VeleroBackup
+	default:
+		return nil, fmt.Errorf("multiple VeleroBackup objects found with label %s=%s in namespace '%s'", constant.NabOriginNameUUIDLabel, labelValue, namespace)
```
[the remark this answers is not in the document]
mpryc: I don't know. It should never happen imo, but I have no strong opinion if it's terminal or not. If it's not the user can remove one object and next reconcile will fix it ?? I think this is really edge case scenario where two same object have labels with same uniue ID.

internal/controller/nonadminbackup_controller_test.go
mpryc: I think it was necessary by the time of writing as I think I was getting some error. Can't remember now.

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -29,63 +29,65 @@ import (
...
+
+	if nonAdminBackup.Status.VeleroBackup != nil {
+		if nonAdminBackup.Status.VeleroBackup.NameUUID == "" {
+			return fmt.Errorf("NonAdminBackup Status VeleroBackupName %v is 0 length string", nonAdminBackup.Status.VeleroBackup.NameUUID)
 		}
-		if !reflect.DeepEqual(nonAdminBackup.Status.VeleroBackup.Status, expectedStatus.VeleroBackup.Status) {
-			return fmt.Errorf("NonAdminBackup Status VeleroBackupStatus %v is not equal to expected %v", nonAdminBackup.Status.VeleroBackup.Status, expectedStatus.VeleroBackup.Status)
+
+		if expectedStatus.VeleroBackup != nil {
+			// When there is no VeleroBackup expected Namespace provided, use one that should be result of reconcile loop
```
mateusoliveira43, line 78: was not easier to understand code with the placeholder approach?

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -259,37 +259,39 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
...
+					Namespace: nonAdminObjectNamespace,
 				},
-				nonAdminBackup,
+				nonAdminBackupAfterReconcile,
 			)).To(gomega.Succeed())
-
-			gomega.Expect(checkTestNonAdminBackupStatus(nonAdminBackup, scenario.ExpectedStatus, nonAdminNamespaceName, oadpNamespaceName)).To(gomega.Succeed())
-
+			gomega.Expect(checkTestNonAdminBackupStatus(nonAdminBackupAfterReconcile, scenario.nonAdminBackupExpectedStatus, oadpNamespace)).To(gomega.Succeed())
+			if scenario.uuidCreatedByReconcile {
```
mateusoliveira43, line 272: this check needs to be moved inside `checkTestNonAdminBackupStatus` function

first expect in this `if` here is not valid, no? if namespace name is big, it wont be present in nameUUID

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -149,107 +148,108 @@ func deleteTestNamespaces(ctx context.Context, nonAdminNamespaceName string, oad
...
 		func(scenario nonAdminBackupSingleReconcileScenario) {
-			updateTestScenario()
-
-			gomega.Expect(createTestNamespaces(ctx, nonAdminNamespaceName, oadpNamespaceName)).To(gomega.Succeed())
-
-			nonAdminBackup := buildTestNonAdminBackup(nonAdminNamespaceName, scenario.spec)
-			gomega.Expect(k8sClient.Create(ctx, nonAdminBackup)).To(gomega.Succeed())
+			nonAdminBackup := buildTestNonAdminBackup(nonAdminObjectNamespace, nonAdminObjectName, scenario.nonAdminBackupSpec)
+			gomega.Expect(k8sClient.Create(ctx, nonAdminBackup.DeepCopy())).To(gomega.Succeed())
+			nonAdminBackupAfterCreate := &nacv1alpha1.NonAdminBackup{}
```
mateusoliveira43, line 203: this get call is unnecessary

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -149,107 +148,108 @@ func deleteTestNamespaces(ctx context.Context, nonAdminNamespaceName string, oad
...
-		updateTestScenario    = func() {
-			counter++
-			nonAdminNamespaceName = fmt.Sprintf("test-nonadminbackup-reconcile-%v", counter)
-			oadpNamespaceName = nonAdminNamespaceName + "-oadp"
-		}
+		ctx                     = context.Background()
+		nonAdminObjectName      string
+		nonAdminObjectNamespace string
+		oadpNamespace           string
+		veleroBackupNameUUID    string
```
mateusoliveira43, line 155: would remove this variable, we only need it in `scenario.uuidFromTestCase == true`, not for all cases

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -465,69 +491,94 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminBackup Controller",
...
+				}
+			})
+			// Check if the context timeout or another error occurred
+			gomega.Expect(err).ToNot(gomega.HaveOccurred(), "Manager failed to start within the timeout period")
 
 			ginkgo.By("Waiting Reconcile of create event")
-			nonAdminBackup := buildTestNonAdminBackup(nonAdminNamespaceName, scenario.spec)
-			gomega.Expect(k8sClient.Create(ctx, nonAdminBackup)).To(gomega.Succeed())
+			nonAdminBackup := buildTestNonAdminBackup(nonAdminObjectNamespace, nonAdminObjectName, scenario.spec)
+			gomega.Expect(k8sClient.Create(ctxTimeout, nonAdminBackup)).To(gomega.Succeed())
```
mateusoliveira43, line 513: is not better to use `ctxTimeout` only in `PollUntilContextTimeout`? my fear is if the test takes more than 10 seconds, it will break

mateusoliveira43 (approved):
note to self: would have done things a little different, setting UUID (only 36 length) as a NAB and Velero Backup label (less secure to use as label instead of status in NAB, but easier to iterate over). After creating UUID, would do get call in NAB and OADP namespace, to confirm UUID is unique in the cluster (https://github.com/migtools/oadp-non-admin/issues/102), but that would not check things in storage provider.

internal/controller/nonadminbackup_controller_test.go
mpryc: The initial implementation had only `time.Sleep()` for `1 second`. It was static sleep expecting the manager will start and be ready after that period of time.

I've changed this to be pulled ever `100 ms` but with limit of `10 seconds`, which is 10x more what it was before.

Assuming that something will be ready after 1 second was not ideal that's why I've added that retry logic here.

internal/controller/nonadminbackup_controller_test.go
mpryc: I prefer to keep all the input or state vars to keep in one place. It's easier to maintain then having test getting some vars from one place and then injecting another from another place.

internal/common/function/function.go
mpryc: I like your implementation better ! Do you want to include in this PR or as follow-up?

internal/common/function/function_test.go
mpryc: They can not, but negative testing is also correct to ensure functions are not broken by some wired change in the code. I think they don't make any harm to the testing.

internal/controller/nonadminbackup_controller_test.go
mpryc: I want the input of the test to be in one place and output in another
Placeholder was introducing some external to reconcile logic which wasn't clear to me and also not always real use case scenario, e.g.:

```
			ExpectedStatus: nacv1alpha1.NonAdminBackupStatus{
				Phase: nacv1alpha1.NonAdminBackupPhaseCreated,
				VeleroBackup: &nacv1alpha1.VeleroBackup{
					Name:      placeholder,
					Namespace: placeholder,
					Status:    &velerov1.BackupStatus{},
				},
```

There is not a chance that expected VeleroBackup will contain placeholder as Name or Namespace if the input VeleroBackup object did not have those fields populated, so we introduced some extra unnecessary logic within the test flow.

internal/controller/nonadminbackup_controller_test.go
mpryc: I don't remember now, but I think it was required, otherwise we were working with modified object after creation which caused wrong data to be compared against.

This one creates the object:
```
gomega.Expect(k8sClient.Create(ctx, nonAdminBackup.DeepCopy())).To(gomega.Succeed())
```

--> reconcile works on the object

---> We need to get object which was after reconcile, to continue working on it.

internal/common/function/function_test.go
mpryc: either way is fine imo

internal/controller/nonadminbackup_controller.go
mpryc: I agree with you that if we carefully check any pre-condition of our reconcile logic as a whole then it is redundant, but having this in place follows a `defensive programming` approach, ensuring that the program doesn't proceed with invalid data, which can prevent wired bugs in the future.

internal/controller/nonadminbackup_controller.go
mpryc: `GetNonAdminLabels()` was meant to be valid for every NAC object also the one which may not require `veleroBackupNameUUID`, that's why I did not want to encorporate this in the function. We can have similar to annotations function:

```
function.GetNonAdminBackupLabels(nab.ObjectMeta)
```

But it will just have that single line of code and we will need to have new function for that, so I felt it's unnecessary addition.

internal/controller/nonadminbackup_controller.go
```diff
@@ -301,32 +341,16 @@ func updateNonAdminPhase(phase *nacv1alpha1.NonAdminBackupPhase, newPhase nacv1a
...
-		status.VeleroBackup.Namespace = veleroBackup.Namespace
-		return true
-	}
-	return false
-}
-
 // updateNonAdminBackupVeleroBackupStatus sets the VeleroBackup status field in NonAdminBackup object status and returns true
 // if the VeleroBackup fields are changed by this call.
 func updateNonAdminBackupVeleroBackupStatus(status *nacv1alpha1.NonAdminBackupStatus, veleroBackup *velerov1.Backup) bool {
+	if status == nil || veleroBackup == nil {
```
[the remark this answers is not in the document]
mpryc: As above re: `defensive programming approach`

internal/common/function/function.go
mpryc: This is generic function which must be on a namespaced object, we don't want to list objects without namespace. We call it inside `GetVeleroBackupByLabel` but we don't check for those values in the `GetVeleroBackupByLabel`. The lowest called function should have check so any caller will benefit from it.

This function in the future will probably be called by sync controller or restore logic, so I think it's not harm to have this check to safeguard.

mpryc (commented):
@mateusoliveira43 I think I responded to all of your comments. Please point me the ones which must be fixed before this will get merged, which we want to fix as follow up and which are just nice to have. Also I am not sure why it's approved by you with so many comments added.

internal/common/function/function.go
mateusoliveira43: either way id fine

mpryc: When replacing with your function I got the test failure. Because current function does work in those test scenarios I will live it as is, because I don't want to spent time on fixing your function:

```
--- FAIL: TestGenerateVeleroObjectNameUUID (0.00s)
    --- FAIL: TestGenerateVeleroObjectNameUUID/nabName_empty (0.00s)
        function_test.go:196: Generated name contains double hyphens: example--e949bb26-67ef-479a-a968-47cebc894f28
    --- FAIL: TestGenerateVeleroObjectNameUUID/very_long_name_and_namespace_empty (0.00s)
        function_test.go:196: Generated name contains double hyphens: rrrrrrrrrrrrrrrrrrrrrrrr--094ab4e3-a9af-42b8-ae4c-33f697052753
    --- FAIL: TestGenerateVeleroObjectNameUUID/empty_namespace_and_empty_name (0.00s)
        function_test.go:196: Generated name contains double hyphens: --5a3d285f-2c2c-4707-9081-fc437f2b299e
```

docs/design/nab_and_nar_status_update.md
mpryc: You are correct we only check if it's empty so removed that part. Also reworded it slightly:

`Does status.VeleroBackup.NameUUID exist ?`

and the if it's not we are going into:


`Generate a NameUUID and set it as status.VeleroBackup.NameUUID` (this step description was: `Set valid status.VeleroBackup.NameUUID`)

internal/controller/nonadminbackup_controller_test.go
mpryc: added back.

docs/design/Non_Admin_Controller_design.md
mpryc: fixed.

internal/controller/nonadminbackup_controller_test.go
mpryc: Removed - ran the tests multiple times and it's not required. Thanks.

internal/controller/nonadminbackup_controller_test.go
mpryc: I don't know how to nicely do that without changing the arguments of the `checkTestNonAdminBackupStatus` function, which I do not want to do at this point. Added TODO and comment in the code.

The problem is that the structure in the `nonAdminBackupSingleReconcileScenario` can not take advantage of variables in an easy way, so we would need to pass some extra args to the `checkTestNonAdminBackupStatus`  instead of using what's in the  `nonAdminBackupSingleReconcileScenario`, bacause we are using here the `nonAdminObjectNamespace` whcih is var `nonAdminObjectNamespace = fmt.Sprintf("test-nab-reconcile-%v", counter)` assigned on every `BeforeEach`

mpryc:
@mateusoliveira43 I've made changes and submitted last commit. There is still one unresolved comment in my opinion, which I've added TODO as I am not sure how to nicely change the current implementation to satisfy your request. Let me know if we can merge it with that or you want to make additional changes.

mateusoliveira43 (approved):

shubham-pampattiwar (approved):

mpryc:
@shubham-pampattiwar 
> /lgtm @mpryc please create the corresponding OADP operator PR for CRD updates. From now on lets do both the PRs is parallel and merge the OADP Operator one first once changes are agreed upon on the NAC PR.

https://github.com/openshift/oadp-operator/pull/1578

Issue #106, opened 2024-10-25
Use apiserver generateName


mateusoliveira43:
## Problem

The `GenerateNacObjectNameWithUUID` function should not be needed

https://github.com/mpryc/oadp-non-admin/blob/velero_nab_ownership/internal/common/function/function.go#L80-L117

We just need UUID (36 char length) for mapping relationships. UUID is stored in NAC objects statuses and Velero objects labels (63 char limit) and as Velero objects names (253 char limit).

## Solution

We should use apiserver generateName functionality instead for Velero objects names

more context: https://github.com/migtools/oadp-non-admin/issues/102

mateusoliveira43:
suggested implementation
```go
func GetGenerateNamePrefix(namespace string, name string) string {
	remainingLength := constant.MaximumNacObjectNameLength - 5 - len("--")
	if len(namespace+name) > remainingLength {
		if len(namespace) >= remainingLength {
			return fmt.Sprintf("%s-", namespace[:remainingLength])
		}
		remainingLength = remainingLength - len(namespace)
		return fmt.Sprintf("%s-%s-", name[:remainingLength], namespace)
	}

	return fmt.Sprintf("%s-%s-", namespace, name)
}
```

Then use it like this
```go
		veleroRestore = &velerov1.Restore{
			ObjectMeta: metav1.ObjectMeta{
				// TODO even though Kubernetes object name can be up to 253 char length
				// using generate name, object name will be 63 char length
				// it will add a random 5 char length to GenerateName prefix
				// if prefix is more than 58 char length, it is truncated
				GenerateName: function.GetGenerateNamePrefix(nar.Namespace, nar.Name),
			},
		}
```

PR #108, opened 2024-10-28
Facilitate deletion of VeleroBackup object via NAB object #58

### LINKED ISSUE #58: Facilitate deletion of VeleroBackup object via NAB object

mpryc:
Fix for deletion of NonAdminBackup object and corresponding VeleroBackup with updated logic to Status not being reconciled every time Status is being changed.

Please check the [docs/design/nab_and_nar_status_update.md diagram](https://github.com/mpryc/oadp-non-admin/blob/velero_delete_event_new/docs/design/nab_and_nar_status_update.md#status-update-scenarios) for the full workflow of the NAB and VeleroBackup deletion and updated Status and Conditions reconciles.
 In few simple steps:
 - The Finalizer is set on NAB object just before VeleroBackup is created
 - The deletion of NAB, can be triggered in three ways:
  1) By setting NAB spec.deleteBackup 
  2) By setting NAB spec.forceDeleteBackup
  3) By running `oc delete` or other k8s API server deletion call such as namespace deletion

## Why the changes were made

To fix #58 

Implemented additional logic within reconcile loop.

This PR replaces #105 

## Testing
1. Deploy OADP with NAC
2. Create NAB Object as in other cases (DPA should also be already in the cluster)
3. Wait for the Velero Backup object to be created
  3a. Run `oc delete <nab>`
        Expected result: NAB not deleted, condition set to inform why it has not been deleted
  3b. With perform steps 1-3a and then set the spec.forceDeleteBackup to true
        Expected result: NAB should be removed, Velero Backup should be removed, S3 Backup should not be removed.
  3c. Perform steps 1-2 and then set the NAB spec.deleteBackup 
        Expected result: NAB should be removed, Velero Backup should be removed, S3 Backup should be removed


api/v1alpha1/nonadminbackup_types.go
```diff
@@ -32,8 +32,8 @@ const (
 	NonAdminBackupPhaseBackingOff NonAdminBackupPhase = "BackingOff"
 	// NonAdminBackupPhaseCreated - Velero Backup was created. The Phase will not have additional informations about the Backup.
 	NonAdminBackupPhaseCreated NonAdminBackupPhase = "Created"
-	// NonAdminBackupPhaseDeletion - Velero Backup is pending deletion. The Phase will not have additional informations about the Backup.
-	NonAdminBackupPhaseDeletion NonAdminBackupPhase = "Deletion"
+	// NonAdminBackupPhaseDeleting - Velero Backup is pending delete. The Phase will not have additional informations about the Backup.
```
kaovilai, line 35:
```suggestion
	// NonAdminBackupPhaseDeleting - Velero Backup is pending deletion. The Phase will not have additional informations about the Backup.
```
or
```suggestion
	// NonAdminBackupPhaseDeleting - Velero Backup is deleting. The Phase will not have additional informations about the Backup.
```

kaovilai (dismissed):

internal/controller/nonadminbackup_controller.go
```diff
@@ -198,11 +480,17 @@ func (r *NonAdminBackupReconciler) validateSpec(ctx context.Context, logger logr
...
+	if nab.Spec.DeleteBackup || !nab.ObjectMeta.DeletionTimestamp.IsZero() {
+		return false, nil
+	}
+
 	if nab.Status.VeleroBackup == nil || nab.Status.VeleroBackup.NameUUID == constant.EmptyString {
 		veleroBackupNameUUID := function.GenerateNacObjectNameWithUUID(nab.Namespace, nab.Name)
 		nab.Status.VeleroBackup = &nacv1alpha1.VeleroBackup{
 			NameUUID:  veleroBackupNameUUID,
 			Namespace: r.OADPNamespace,
+			Name:      veleroBackupNameUUID,
```
mateusoliveira43, line 493: why duplicate the information?

internal/controller/nonadminbackup_controller.go
```diff
@@ -113,7 +385,12 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 // If it is empty, it sets the Phase to "New".
 // It then returns boolean values indicating whether the reconciliation loop should requeue or exit
 // and error value whether the status was updated successfully.
-func (r *NonAdminBackupReconciler) init(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+func (r *NonAdminBackupReconciler) initNabCreate(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	// If the NonAdminBackup is marked for deletion (either by DeleteBackup in the spec or with a DeletionTimestamp), skip initNabCreate
+	if nab.Spec.DeleteBackup || !nab.ObjectMeta.DeletionTimestamp.IsZero() {
```
mateusoliveira43, line 390: please remove all this checks in the following functions

Reconcile should not reach this stage of reconcile if it is marked for deletion

internal/controller/nonadminbackup_controller.go
mpryc: It is to keep consistency between Status fields for `VeleroBackup` and `VeleroDeleteBackupRequest`. We do not guarantee that Name is the key that links NAB with Velero. The field `NameUUID` is to be used as reference only, for example:

https://github.com/migtools/oadp-non-admin/pull/108/files/289367e71f9939098a18213c5205d6fa5a50cee6#diff-1fc992498fb761a2b0d175e637f5f945c2eb932ffacb0713d8724f4c14054c31R87

In the `VeleroBackup` we use the NameUUID, however in the `VeleroDeleteBackupRequest` we use name generated similarly how Velero would gneerate `DeleteBackupRequest` name, so we need to have both fields there. In case of `VeleroBackup` those fields will be the same, however as I wrote for consistency it's good in my opinion to have similar structure of the fields.

internal/controller/nonadminbackup_controller.go
mpryc: But this will continue to enter those functions, right?

we do not have `EXIT NOW RECONCILE` return statement, so having `return false, nil` will continue to the next reconcile function and if it's marked as delete it will enter them?

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+// This function first checks if the NAB object has the finalizer. If yes, it attempts to locate the associated
+// VeleroBackup object by the UUID stored in the NAB status. If a unique VeleroBackup object is found,
+// deletion is initiated, and the reconcile loop will be requeued. If multiple VeleroBackup objects
+// are found with the same label, the function logs an error and returns.
+//
+// Returns:
+//   - A boolean indicating whether to requeue the reconcile loop,
+//   - An error if VeleroBackup deletion or retrieval fails, for example, due to multiple VeleroBackup objects found with the same UUID label.
+func (r *NonAdminBackupReconciler) createVeleroDeleteBackupRequest(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	if !controllerutil.ContainsFinalizer(nab, constant.NabFinalizerName) ||
```
mateusoliveira43, line 245: I think this is not needed

this is only called after `setStatusAndConditionForDeletionAndCallDelete` function. If this function is reached, we know NAB has finalizer (and UUID)

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+					velerov1.BackupUIDLabel, string(veleroBackup.UID),
+					constant.NabOriginNameUUIDLabel, veleroBackupNameUUID,
+				),
+				builder.WithLabelsMap(function.GetNonAdminLabels()),
+				builder.WithAnnotationsMap(function.GetNonAdminBackupAnnotations(nab.ObjectMeta)),
+				builder.WithGenerateName(veleroBackup.Name+"-"),
+			).Result()
+
+		// Use CreateRetryGenerateName for retry logic in creating the delete request
+		if err := function.CreateRetryGenerateName(r.Client, ctx, deleteBackupRequest); err != nil {
```
mateusoliveira43, line 290: I think we should not create function for this

Using this https://github.com/migtools/oadp-non-admin/issues/106 would solve the problem here, no?

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+// Parameters:
+//   - ctx: Context for managing request lifetime
+//   - logger: Logger instance
+//   - nab: NonAdminBackup object
+//
+// Returns:
+//   - bool: whether to requeue (always false)
+//   - error: any error encountered during deletion
+func (r *NonAdminBackupReconciler) deleteVeleroBackupAndDeleteBackupRequestObjects(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	if nab.Status.VeleroBackup == nil || nab.Status.VeleroBackup.NameUUID == constant.EmptyString {
```
mateusoliveira43, line 327: would also remove this `if`

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if err = r.Delete(ctx, veleroBackup); err != nil {
+			logger.Error(err, "Failed to delete VeleroBackup", nameString, veleroBackup.Name)
+			return false, err
+		}
+		logger.V(1).Info("VeleroBackup deletion initiated", nameString, veleroBackup.Name)
+	} else {
+		logger.V(1).Info("VeleroBackup already deleted")
+	}
+
+	deleteBackupRequest, err := function.GetVeleroDeleteBackupRequestByLabel(ctx, r.Client, r.OADPNamespace, veleroBackupNameUUID)
```
mateusoliveira43, line 351: since DeleteBackupRequest controller will, eventually, do this clean up for us, would let it do the deletion for us

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+// After confirming VeleroBackup deletion, it removes the finalizer from the NAB, allowing Kubernetes
+// to complete the garbage collection process for the NAB object itself.
+//
+// Returns:
+//   - A boolean indicating whether to requeue the reconcile loop (true if waiting for VeleroBackup deletion).
+//   - An error if any update operation or deletion check fails.
+func (r *NonAdminBackupReconciler) removeNabFinalizerUponVeleroBackupDeletion(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	// Check if DeleteBackup is set or NAB was deleted
+	if !nab.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !nab.Spec.ForceDeleteBackup && nab.Status.VeleroBackup != nil && nab.Status.VeleroBackup.NameUUID != constant.EmptyString {
```
mateusoliveira43, line 388: would also remove these `if`s

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+// Parameters:
+//   - ctx: Context for managing request lifetime
+//   - logger: Logger instance
+//   - nab: NonAdminBackup being deleted
+//
+// Returns:
+//   - bool: true if status was updated
+//   - error: any error encountered
+func (r *NonAdminBackupReconciler) setStatusForDirectKubernetesApiDeletion(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	if controllerutil.ContainsFinalizer(nab, constant.NabFinalizerName) {
```
mateusoliveira43, line 205: this function is only called if deletion timestamp is greater than zero, so that implies finalizer exists, right? would remove `if`

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				Message: "backup deletion requires VeleroBackup deletion or finalizer removal",
+			},
+		)
+		if updatedPhase || updatedCondition {
+			if err := r.Status().Update(ctx, nab); err != nil {
+				logger.Error(err, statusUpdateError)
+				return false, err
+			}
+			logger.V(1).Info("NonAdminBackup status marked for deletion")
+			return true, nil
```
mateusoliveira43, line 221: here we can exit, no? If it enters reconcile again, it does not have anything else to do

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			ObjectMeta(
+				builder.WithLabels(
+					velerov1.BackupNameLabel, label.GetValidName(veleroBackup.Name),
+					velerov1.BackupUIDLabel, string(veleroBackup.UID),
+					constant.NabOriginNameUUIDLabel, veleroBackupNameUUID,
+				),
+				builder.WithLabelsMap(function.GetNonAdminLabels()),
+				builder.WithAnnotationsMap(function.GetNonAdminBackupAnnotations(nab.ObjectMeta)),
+				builder.WithGenerateName(veleroBackup.Name+"-"),
+			).Result()
```
mateusoliveira43, line 287: It not easier to read if we create the YAML direct?

something like
```go
		deleteBackupRequest = &velerov1.DeleteBackupRequest{
			ObjectMeta: metav1.ObjectMeta{
				Namespace:    r.OADPNamespace,
				Labels:       function.GetNonAdminLabels(),
				Annotations:  function.GetNonAdminBackupAnnotations(nab.ObjectMeta),
				GenerateName: veleroBackup.Name + "-",
			},
			Spec: velerov1.DeleteBackupRequestSpec{
				BackupName: veleroBackup.Name,
			},
		}
```

internal/controller/nonadminbackup_controller.go
mpryc: No, it get's called on any Kubernetes API `delete` call regardless if the NAB contains finalizer or not, so we do not want to set status on the object that does not have finalizer, we just want to proceed with API call which will make object to be gone.

api/v1alpha1/nonadminbackup_types.go
```diff
@@ -43,6 +45,18 @@ type NonAdminBackupSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=trace;debug;info;warning;error;fatal;panic
 	LogLevel string `json:"logLevel,omitempty"`
+
+	// DeleteBackup tells the controller to remove created Velero Backup
```
mateusoliveira43, line 49: I think is important to doc that this deletes storage data as well

internal/controller/nonadminbackup_controller.go
mpryc: I really wanted to make this as similar to what Velero CLI does as possible so we don't introduce the `DeleteBackupRequest` that is unique to NAC only. We are taking what's in the Velero and adding on top our annotations and labels. Please check how Velero CLI is creating such object:
https://github.com/vmware-tanzu/velero/blob/6bffac5d0661ad066ea487774b3862b56067507d/pkg/cmd/cli/backup/delete.go#L124

internal/controller/nonadminbackup_controller.go
mpryc: It's exactly how Velero CLI is doing, so I want this part to be very similar. If you go to the function.CreateRetryGenerateName you will see comment pointing to the Velero part from which the logic was borrowed.

internal/controller/nonadminbackup_controller.go
mpryc: @shubham-pampattiwar please comment on this.

internal/controller/nonadminbackup_controller.go
mateusoliveira43: I think this retry is not needed (if create fails, reconcile exits with error and then it will enter the reconcile again and retry creation here)

but if you want to use this function, then delete duplication from `internal/common/function/function.go`, just import from velero package directly

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+		// Use CreateRetryGenerateName for retry logic in creating the delete request
+		if err := function.CreateRetryGenerateName(r.Client, ctx, deleteBackupRequest); err != nil {
+			logger.Error(err, "Failed to create delete request for VeleroBackup", "BackupName", veleroBackup.Name)
+			return false, err
+		}
+		logger.V(1).Info("Request to delete backup submitted successfully", "BackupName", veleroBackup.Name)
+		nab.Status.VeleroDeleteBackupRequest = &nacv1alpha1.VeleroDeleteBackupRequest{
+			NameUUID:  veleroBackupNameUUID,
```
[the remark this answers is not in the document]
mpryc: But if you have 3 objects:
  A - where the generated `nameuuid` first appears
  B - we have the link to the `A` object by using `nameuuid`
  C - we have the link to the `A` object by using `nameuuid`

I don't know how we could link `C` or `B` to `A` if we delete those identifiers from both or even from one of `B` or `C` ?

internal/controller/nonadminbackup_controller.go
```diff
@@ -261,6 +599,11 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
 		// situations where NAC object do not require NabOriginUUIDLabel
 		veleroBackup.Labels[constant.NabOriginNameUUIDLabel] = veleroBackupNameUUID
 
+		// Skip the createVeleroBackupAndSyncWithNonAdminBackup, due to finalizer missing
```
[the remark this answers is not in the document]
mpryc: agree

internal/controller/nonadminbackup_controller.go
mpryc: If we do and for some reason those are nil, we have a controller panic. I can easily replicate this behavior by not enabling velero controller and creating delete request on NAB. This way nab.StatusVeleroBackup is never set and we got panic here.

internal/controller/nonadminbackup_controller.go
mpryc: It does not, but I want to keep it same across all status updates, so whenever it succeeded or not we do want to either exit with error or reconcile so next time it enters reconciles there is no update and ends the reconcile.

This way we have same pattern for each status update, otherwise we may be introducing variations which may cause later some wired behavior.

@shubham-pampattiwar any opinion on that?

api/v1alpha1/nonadminbackup_types.go
mpryc: `DeleteBackup tells the controller to create DeleteBackupRequest which removes all backup data including backup storage and Velero Backup objects`

Is the above fine?

internal/controller/nonadminbackup_controller.go
mpryc: I agree with you. Will remove that check.

api/v1alpha1/nonadminbackup_types.go
mateusoliveira43: I would not tell about implementation detail (DeleteBackupRequest) to user, but it is ok, the message contains the important part

internal/controller/nonadminbackup_controller.go
mpryc: Ok I removed that and added comment in the code, why.

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		logger.V(1).Info("NonAdminBackup status marked for deletion")
+		return true, nil
+	}
+	if nab.ObjectMeta.DeletionTimestamp.IsZero() {
+		logger.V(1).Info("Marking NonAdminBackup for deletion", nameString, nab.Name)
+		if err := r.Delete(ctx, nab); err != nil {
+			logger.Error(err, "Failed to call Delete on the NonAdminBackup object")
+			return false, err
+		}
+		return true, nil // Requeue to allow deletion to proceed
```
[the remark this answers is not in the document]
mpryc: We can think differently:
 - if update, we requeue skipping the predicate logic and save some resources on that.
 - if object is deleted we skip predicate and our logic will exit as well on:

So in both cases we save on predicate logic if we know the re-queue is needed to proceed
 
```
	err := r.Get(ctx, req.NamespacedName, nab)
	if err != nil {
		if apierrors.IsNotFound(err) {
```

api/v1alpha1/nonadminbackup_types.go
mpryc: Ok will reword this to be aligned with https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/backup-reference.md#deleting-backups

internal/controller/nonadminbackup_controller.go
mpryc: ok will use that from velero directly.

internal/controller/nonadminbackup_controller.go
```diff
@@ -56,6 +60,10 @@ const (
 	veleroReferenceUpdateRequeue = "NonAdminBackup - Requeue after Status Update with UUID reference"
 	statusUpdateExit             = "NonAdminBackup - Exit after Status Update"
 	statusUpdateError            = "Failed to update NonAdminBackup Status"
+	findSingleVBError            = "Failed to find single VeleroBackup object"
```
shubham-pampattiwar, line 63: you need to change the err string here. This might imply that you could not find a single VB for the particular UUID. I would reframe it to some thing like: `Error encountered while retrieving VeleroBackup for NAB during the Delete operation` your get function already has the particular error.

internal/controller/nonadminbackup_controller.go
```diff
@@ -56,6 +60,10 @@ const (
...
+	findSingleVDBRError          = "Failed to find single DeleteBackupRequest object"
```
shubham-pampattiwar, line 64: Similar suggestion as above.

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+func (r *NonAdminBackupReconciler) createVeleroDeleteBackupRequest(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	if !controllerutil.ContainsFinalizer(nab, constant.NabFinalizerName) ||
+		nab.Status.VeleroBackup == nil ||
+		nab.Status.VeleroBackup.NameUUID == constant.EmptyString {
+		return false, nil
+	}
+
+	// Initiate deletion of the VeleroBackup object only when the finalizer exists.
+	// For the ForceDelete we do not create DeleteBackupRequest
+	veleroBackupNameUUID := nab.Status.VeleroBackup.NameUUID
```
shubham-pampattiwar, line 253: Lets call it NACUUID

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+					constant.NabOriginNameUUIDLabel, veleroBackupNameUUID,
+				),
+				builder.WithLabelsMap(function.GetNonAdminLabels()),
+				builder.WithAnnotationsMap(function.GetNonAdminBackupAnnotations(nab.ObjectMeta)),
+				builder.WithGenerateName(veleroBackup.Name+"-"),
+			).Result()
+
+		// Use CreateRetryGenerateName for retry logic in creating the delete request
+		if err := function.CreateRetryGenerateName(r.Client, ctx, deleteBackupRequest); err != nil {
+			logger.Error(err, "Failed to create delete request for VeleroBackup", "BackupName", veleroBackup.Name)
```
shubham-pampattiwar, line 291: I think we need to add NAB name here ? I am good with adding VB as well as NAB name in the log.

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return false, err
+		}
+		logger.V(1).Info("Request to delete backup submitted successfully", "BackupName", veleroBackup.Name)
```
shubham-pampattiwar, line 294: Similarly add NAB name here as well, as user interacts with NAB and not VB.

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,279 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+			veleroBackup, err := function.GetVeleroBackupByLabel(ctx, r.Client, r.OADPNamespace, veleroBackupNameUUID)
+			if err != nil {
+				// Case in which more then one VeleroBackup is found with the same label UUID
+				// TODO: Should we delete all of the objects with such UUID ?
+				logger.Error(err, findSingleVBError, uuidString, veleroBackupNameUUID)
+				return false, err
+			}
+
+			if veleroBackup != nil {
```
shubham-pampattiwar, line 399: Hmm, this might not be sufficient, right ? We might also need to check up on DeleteBackupRequest associated with the VB ? The status of DBR might give more details to us and then based on that we can go further in the deletion workflow, WDYT ?

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: yeah we could directly re-use from velero.

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: Yeah Velero eventually cleans up the residual DBR. But there is no harm in deleting the DRBs ourselves right ? I think its a good step to cleanup DBR objects that we created for the particular NAB and VB and not keep them dangling in cluster.

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: yeah i think its time to rename the UUID field as I mentioned in the other comment.

internal/controller/nonadminbackup_controller.go
```diff
@@ -83,12 +91,51 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-	reconcileSteps := []reconcileStepFunction{
-		r.init,
-		r.validateSpec,
-		r.setBackupUUIDInStatus,
-		r.createVeleroBackupAndSyncWithNonAdminBackup,
+	// Determine which path to take
+	var reconcileSteps []reconcileStepFunction
+
+	// First switch statement takes precedence over the next one
```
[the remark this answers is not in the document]
mateusoliveira43: to simplify, I would let user specify both at the same time. We can doc that force takes precedence 

I believe use case would be "I told it to delete, its taking too long, I will force delete then"

api/v1alpha1/nonadminbackup_types.go
```diff
@@ -55,16 +69,42 @@ type VeleroBackup struct {
...
 
+// VeleroDeleteBackupRequest contains information of the related Velero delete backup request object.
+type VeleroDeleteBackupRequest struct {
+	// status captures the current status of the Velero delete backup request.
+	// +optional
+	Status *velerov1.DeleteBackupRequestStatus `json:"status,omitempty"`
+
+	// nameuuid references the Velero delete backup request object by it's label containing same NameUUID.
+	// +optional
+	NameUUID string `json:"nameuuid,omitempty"`
```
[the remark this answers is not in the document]
mpryc: renaming to `NACUUID`

internal/common/function/function.go
```diff
@@ -35,6 +38,21 @@ import (
 	"github.com/migtools/oadp-non-admin/internal/common/constant"
 )
 
+// CreateRetryGenerateName attempts to create an object, retrying on "AlreadyExists" errors if the object’s name is generated.
+// borrowed from https://github.com/vmware-tanzu/velero/blob/8320df44fda00daf802b60a42ba2327bb5f39e15/pkg/client/retry.go#L30
+func CreateRetryGenerateName(k8sClient client.Client, ctx context.Context, obj client.Object) error {
```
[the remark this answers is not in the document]
mpryc: The intention was exactly like this, to reuse that function and not rely on eventual changes in velero cli, but as in the comment I am now replacing it with direct use from veler lib:
https://github.com/migtools/oadp-non-admin/pull/108#discussion_r1830580177

internal/controller/nonadminbackup_controller.go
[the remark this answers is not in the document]
mpryc: I think it's out of scope for this PR. The validation as such should be done in my opinion via validating webhook.

internal/controller/nonadminbackup_controller.go
```diff
@@ -83,12 +91,51 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	case nab.Spec.DeleteBackup:
+		// Standard delete path - creates DeleteBackupRequest and waits for VeleroBackup deletion
+		logger.V(1).Info("Executing standard delete path")
+		reconcileSteps = []reconcileStepFunction{
+			r.setStatusAndConditionForDeletionAndCallDelete,
+			r.createVeleroDeleteBackupRequest,
+			r.removeNabFinalizerUponVeleroBackupDeletion,
+		}
+
+	case !nab.ObjectMeta.DeletionTimestamp.IsZero():
```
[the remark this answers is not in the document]
mpryc: @shubham-pampattiwar exactly as you wrote.

The condition which I am currently setting tells the user what are possible actions and why the NAB is hanging there. The object will have condition:

```
Condition:    Deleting
Reason:       DeletionPending
Message:      backup deletion requires VeleroBackup deletion or finalizer removal
```

The real idea was to tell user that an extra action is required and leave the decision up to the user what to do with that NAB object. We could also include message telling that the real data is not removed from s3 storage at this point and requires setting spec field to complete. I see people do not fully understand that `oc delete backup` won't  delete real backup, hence in OADP docs we have:
   https://docs.openshift.com/container-platform/4.17/backup_and_restore/application_backup_and_restore/backing_up_and_restoring/oadp-deleting-backups-doc.html

internal/controller/nonadminbackup_controller.go
mpryc: Note this is different then https://github.com/migtools/oadp-non-admin/pull/108#discussion_r1830585938 because in this path we are certain DBR was removed. In the https://github.com/migtools/oadp-non-admin/pull/108#discussion_r1830585938 we do want to clean this without checking why that failed (force).

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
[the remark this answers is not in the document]
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
mpryc: Used directly from velero in the newest commit.

internal/controller/nonadminbackup_controller.go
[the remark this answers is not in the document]
mpryc: Please respond if we need to rework this part.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

api/v1alpha1/nonadminbackup_types.go
mpryc: Done please check newest commit.

api/v1alpha1/nonadminbackup_types.go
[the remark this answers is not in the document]
mpryc: Done please check newest commit.

internal/common/function/function.go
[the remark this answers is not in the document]
mpryc: The function however is used now directly from velero. Resolving this conversation as there is similar open above.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

internal/controller/nonadminbackup_controller.go
mpryc: Done please check newest commit.

deepsm007:
/test images

infra issue

internal/controller/nonadminbackup_controller.go
```diff
@@ -97,11 +144,277 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+//   - bool: true if status was updated
+//   - error: any error encountered
+func (r *NonAdminBackupReconciler) setStatusForDirectKubernetesAPIDeletion(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	updatedPhase := updateNonAdminPhase(&nab.Status.Phase, nacv1alpha1.NonAdminBackupPhaseDeleting)
+	updatedCondition := meta.SetStatusCondition(&nab.Status.Conditions,
+		metav1.Condition{
+			Type:    string(nacv1alpha1.NonAdminConditionDeleting),
+			Status:  metav1.ConditionTrue,
+			Reason:  "DeletionPending",
+			Message: "backup deletion requires VeleroBackup deletion or finalizer removal",
```
mateusoliveira43, line 211: here message is misleading, not?

If I do `oc delete` and deletes Velero backup, NAB will not be deleted

what about something like: 
> backup deletion requires setting spec.deleteBackup or spec.forceDeleteBackup to true or finalizer removal

internal/controller/nonadminbackup_controller.go
mateusoliveira43: agree with Michal here, I think current check is enough

mpryc:
The last update to the reconcile and I've created Issue to track next PR:
https://github.com/migtools/oadp-non-admin/issues/109

internal/controller/nonadminbackup_controller.go
mpryc: Fixed.

mpryc:
@shubham-pampattiwar @mateusoliveira43 in my opinion it's ready for final review. Updated design diagrams (actually completely reworked them) to be easier to read, added tests and addressed those comments that I've decided are important.

internal/controller/nonadminbackup_controller.go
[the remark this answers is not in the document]
shubham-pampattiwar: Resolving this conversation, created an issue for this topic https://github.com/migtools/oadp-non-admin/issues/112

mateusoliveira43 (changes requested):
Update PR description with how to test, link OADP PR and telling about changes of the PR (there is no mention to `forceDeleteBackup` or the changes made to other parts of the code that do not involve deletion)

config/rbac/role.yaml
```diff
@@ -36,6 +36,19 @@ rules:
   - backups
   verbs:
   - create
+  - delete
```
shubham-pampattiwar, line 39: We need this updated in OADP CSV as well. Please update the corresponding OADP Operator PR: https://github.com/openshift/oadp-operator/pull/1588

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -543,13 +861,13 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminBackup Controller",
...
 					types.NamespacedName{
-						Name:      nonAdminBackup.Status.VeleroBackup.NameUUID,
+						Name:      nonAdminBackup.Status.VeleroBackup.NACUUID,
 						Namespace: oadpNamespace,
 					},
 					veleroBackup,
 				)).To(gomega.Succeed())
 
-				// can not call .Status().Update() for veleroBackup object https://github.com/vmware-tanzu/velero/issues/8285
+				// TODO can not call .Status().Update() for veleroBackup object: backups.velero.io "name..." not found error
```
mateusoliveira43, line 870: undo this comment change

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -238,6 +248,14 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
 				}
 				gomega.Expect(k8sClient.Create(ctx, veleroBackup)).To(gomega.Succeed())
 			}
+
+			// We allow to have deleteNonAdminBackup and forceDeleteNonAdminBackup set to true at the same time
+			if scenario.nonAdminBackupSpec.ForceDeleteBackup {
```
mateusoliveira43, line 253: what you want to do with this? Add DeletionTimestamp to some tests?

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -260,53 +278,345 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
...
+					{
+						Type:               string(nacv1alpha1.NonAdminConditionDeleting),
+						Status:             metav1.ConditionTrue,
+						Reason:             "DeletionPending",
+						Message:            "backup accepted for deletion",
+						LastTransitionTime: metav1.NewTime(time.Now()),
+					},
+				},
+			},
+			nonAdminBackupExpectedStatus: nacv1alpha1.NonAdminBackupStatus{
```
mateusoliveira43, line 565: remove this, is not used by test

internal/controller/nonadminbackup_controller_test.go
mpryc: Some of the reconcile logic is based on the `DeletionTimestamp.IsZero()` so I want to ensure that object is pending deletion to test just part responsible for it (one function instead of entire switch part)

mpryc: here is example of failure of the test without this:
```shell
••••••••
------------------------------
• [FAILED] [0.014 seconds]
Test single reconciles of NonAdminBackup Reconcile function Reconcile triggered by NonAdminBackup Create/Update events and by Requeue [It] When triggered by NonAdminBackup forceDeleteNonAdmin spec field with Finalizer set, should delete NonAdminBackup and exit
./internal/controller/nonadminbackup_controller_test.go:526

  Timeline >>
  2024-11-13T17:06:54+01:00	DEBUG	NonAdminBackup Reconcile start
  2024-11-13T17:06:54+01:00	DEBUG	Executing force delete path
  2024-11-13T17:06:54+01:00	DEBUG	NonAdminBackup status unchanged during deletion
  2024-11-13T17:06:54+01:00	DEBUG	Marking NonAdminBackup for deletion	{"name": "nab-object-8"}
  [FAILED] in [It] - ./internal/controller/nonadminbackup_controller_test.go:266 @ 11/13/24 17:06:54.943
  << Timeline

  [FAILED] Expected
      <reconcile.Result>: {Requeue: true, RequeueAfter: 0}
  to equal
      <reconcile.Result>: {Requeue: false, RequeueAfter: 0}
  In [It] at: ./internal/controller/nonadminbackup_controller_test.go:266 @ 11/13/24 17:06:54.943
------------------------------
•••••


```

internal/controller/nonadminbackup_controller_test.go
mpryc: done

internal/controller/nonadminbackup_controller_test.go
mateusoliveira43: but you are doing it based on nonAdmin object, not the test. I think this is wrong

correct approach would be creating flag like `addDeletionTimestamp` and passing as true to that test

mateusoliveira43:
/override "security/snyk ( Hybrid Platforms - OpenShift API for Data Protection)"

internal/controller/nonadminbackup_controller_test.go
mpryc: yes, however in this scenario we don't require another flag because there is no case where we will use this without force delete. Is to basically go into ForceDelete switch statement which is controlled by that `DeletionTimestamp.IsZero()` flag in the https://github.com/migtools/oadp-non-admin/pull/108/files#diff-98b8a647a66c83f5ee56f93b7530fb37a28301390936ecdb1fdaf38c031453c6R116

mpryc: Added `addNabDeletionTimestamp` as test flag.

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -481,7 +482,8 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
 			result:                        reconcile.Result{Requeue: true},
 		}),
 		ginkgo.Entry("When triggered by NonAdminBackup forceDeleteNonAdmin spec field with Finalizer set and NonAdminBackup phase Created without DeletionTimestamp, should trigger delete NonAdminBackup and requeue", nonAdminBackupSingleReconcileScenario{
-			addFinalizer: true,
+			addFinalizer:            true,
+			addNabDeletionTimestamp: true,
```
mateusoliveira43, line 486: test name says `without DeletionTimestamp`. Why is true here?

mpryc: You are right, I've modified this to false, however this test case is without deletion timestamp as it never get's into that reconcile logic. Following test is to test from the point where `DeletionPending` condition is set. This one until that condition. force push the small nit to upde this.

mateusoliveira43:
/override "security/snyk ( Hybrid Platforms - OpenShift API for Data Protection)"

shubham-pampattiwar (approved):

Issue #57, opened 2024-05-02
Non-Admin BSL credential design


weshayutin:
* Non-Admin users are will be able to create BSL's 
* Ideally the DPA is not changed when a non-admin BSL is created

As discussed in scrum on 5/2
* A non-admin bsl will be reconciled by the non-admin controller
* The secret used from the non-admin bsl will be copied into a single credential in the oadp namespace
* An empty secret in oadp namespace will be created when non-admin is enabled and mounted to the velero container
  * Perhaps a side car can be used and restarted in lieu of the velero pod 
* non-admin creds will need to be namespaced in the cred key.

mpryc:
The design is part of the #139. Please let me know if this is sufficient or some clarification is needed. It can be viewed in the:
https://github.com/migtools/oadp-non-admin/blob/6223d3700392f3f39bf439ca860b096cc28ffcdd/docs/design/non_admin_backupstoragelocation.md

shubham-pampattiwar:
Closing as completed https://github.com/migtools/oadp-non-admin/pull/139

PR #115, opened 2024-11-18
Allow Non-Admin user to create Backup Storage Locations - fixes #36

### LINKED ISSUE #36: Ability for non-admin users to create BackupStorageLocations(BSL)
Add the the functionality of providing the non-admin users the ability to create their own BackupStorageLocations. In other words BYOB (Bring your own Bucket/BSL). The task would entail:
- Introduction of a Non-Admin NackupStorageLocation CRD (NABSL)
- NABSL controller 
    - The NABSL controller would cascade the BSL request to Velero controller
    - Validation that the NABSL CR is appropriate and relevant secrets from non-admin user are also obtained (note that authenticating the access keys will not be the NABSL controller's responsiblity)
    - The NABSL request can be create/update/view/delete BSL type requests
    - NABSL controller would be responsible of gathering the status Velero BSL and updating the status of NABSL CR, keeping them in sync 
    - Every BSL needs user access keys and storage credentials, NABSL controller will be responsible for provisioning them in Velero NS, once they are provided by the non-admin user
- Addition of Validating Webhooks to ensure that only give Velero CLI/OC access is available to the non-admin user's relevant BSL CRs 
- Control over general Velero BSL Spec that gets exposed or allow listed via NABSL CRD spec.

mpryc:
## Why the changes were made

Fix #36, also #57

1. Create as usual OADP with NAC.
2. Create user secret in the user's namespace similarly to the S3 secret for DPA:

```shell
$ oc create secret generic nac-credentials -n <USERS_NAMESPACE> --from-file cloud=credentials-velero
```
3. Create Non Admin BSL pointing to the above credential in the user's namespace, example spec part:
```yaml

metadata:
  name: example
  namespace: nactest
spec:
  config:
    region: eu-central-1
  credential:
    key: cloud
    name: nac-credentials
  objectStorage:
    bucket: my-bucket
    prefix: some-prefix
  provider: aws

```

4. Check the NAC logs, the NaBSL object (if it was correctly reconciled and updated with status from Velero BSL)

5...6...7... Do break things, such as improper Secret data, non-existing secret, removal of the secret from the OADP namespace or the user namespace, removal of the BSL from the OADP namespace and see if everything reconciles as expected.

Read [design](https://github.com/mpryc/oadp-non-admin/blob/nonadminstoragelocation_controller/docs/design/non_admin_backupstoragelocation.md)


mpryc:
@mateusoliveira43 @shubham-pampattiwar @weshayutin enjoy playing with it, comments welcome.
Note that tests will be implemented after we agree on the current implementation. This part is for creation/deletion of BSL. Actual use of the BSL from the Non Admin Backup should be in separate PR as it will require some changes in the non-admin Backup reconcile to ensure we are safe-guarding against usage from BSL's which are not allowed for the user.

shawn-hurley:
What happens if the BSL already exists in the velero namespace?

api/v1alpha1/nonadminbackupstoragelocation_types.go
```diff
@@ -0,0 +1,107 @@
...
+*/
+
+package v1alpha1
+
+import (
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
mateusoliveira43, line 24: remove scaffolded comments

cmd/main.go
```diff
@@ -30,8 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
```
mateusoliveira43, line 33: undo

config/rbac/kustomization.yaml
```diff
@@ -16,3 +16,9 @@ resources:
 # - auth_proxy_role.yaml
 # - auth_proxy_role_binding.yaml
 # - auth_proxy_client_clusterrole.yaml
+# For each CRD, "Editor" and "Viewer" roles are scaffolded by
```
mateusoliveira43, line 19: did you use correct kubebuilder version https://github.com/migtools/oadp-non-admin/blob/ca7e290f36a116593ba8255125baf3b8b561dc6c/docs/architecture.md#kubebuilder ? I did not expect this change

mateusoliveira43 (changes requested):
need to update architecture commands https://github.com/migtools/oadp-non-admin/blob/ca7e290f36a116593ba8255125baf3b8b561dc6c/docs/architecture.md#kubebuilder

config/rbac/kustomization.yaml
mpryc: The scaffold was using operator-sdk with the following versions:

```shell
# operator-sdk version
operator-sdk version: "v1.37.0", commit: "819984d4c1a51c8ff2ef6c23944554148ace0752", kubernetes version: "1.29.0", go version: "go1.21.13", GOOS: "linux", GOARCH: "amd64"

# kubebuilder version
Version: main.version{KubeBuilderVersion:"4.2.0", KubernetesVendor:"1.31.0", GitCommit:"c7cde5172dc8271267dbf2899e65ef6f9d30f91e", BuildDate:"2024-08-17T09:41:45Z", GoOs:"linux", GoArch:"amd64"}

```

See the commit message:
  https://github.com/migtools/oadp-non-admin/pull/115/commits/6d9db57d78a263be9cc1cc1b07a66d4d93e7b210

mateusoliveira43: this project does not use operator-sdk

mpryc: The generated api is same as under the hood operator-sdk calls kubebuilder, question is if I should use older kubebuilder version instead of newer one?

mpryc: I am currently rebasing this on top of kubebuilder generated scaffold with the same version 1.34.0. Maybe it makes sense to use same version, so later we can upgrade to newer one from same base.

mpryc:
I am closing this PR in favor of https://github.com/migtools/oadp-non-admin/pull/139

PR #117, opened 2024-11-21
Added Estimated Queue Position in the Status field of the NAB object

### LINKED ISSUE #20: Inform non admin user how much operations are left before running its request
Since Velero does not process backups in parallel (one at a time, in order of creation timestamp), non admin users need information about how much backups (without time estimation) are queued before running the requested NonAdminBackup. 

This information can be shown to admin users by running `velero backup get`, but since non admin users do not know about other non admin users backups (and admin user backups), it is important to inform this in status of NonAdminBackup object. This will be done by adding:
- a condition to represent it is queued
- and how many are queued:
  - 0 meaning it is currently running 
  - greater than 0 meaning that that amount of backups need to complete before running requested NonAdminBackup.

Same thing applies to restores and  NonAdminRestores.

mpryc:
## Why the changes were made

To fix #20. Included in this fix is also small fix for the #119.

There were unit and functional tests written to cover newly written functions within `function.go`.
There was `estimatedQueuePosition` added to the tests of the reconcile to include that coverage. Note no new tests in the `nonadminbackup_controller_test.go` were added, because it was straight forward to add test coverage to already existing test cases as part of the status check.

1. Written tests, please run:
```shell
$ make test
[...]
	github.com/migtools/oadp-non-admin/api/v1alpha1		coverage: 0.0% of statements
?   	github.com/migtools/oadp-non-admin/internal/common/constant	[no test files]
	github.com/migtools/oadp-non-admin/cmd		coverage: 0.0% of statements
ok  	github.com/migtools/oadp-non-admin/internal/common/function	0.018s	coverage: 50.3% of statements
	github.com/migtools/oadp-non-admin/internal/handler		coverage: 0.0% of statements
	github.com/migtools/oadp-non-admin/test/utils		coverage: 0.0% of statements
	github.com/migtools/oadp-non-admin/internal/predicate		coverage: 0.0% of statements
ok  	github.com/migtools/oadp-non-admin/internal/controller	14.405s	coverage: 58.2% of statements
```

2. To test the functionality on the test cluster:

- Create deployment as usual with the OADP and NonAdminBackup controllers
- Create non-admin namespace
- One could run the following shell script to create 20 NAB objects by setting `$ export BACKUP_NAMESPACE=my-non-admin-namespace`:
```shell
#!/bin/bash

if [ -z "$BACKUP_NAMESPACE" ]; then
  echo "ERROR: BACKUP_NAMESPACE environment variable is not set."
  exit 1
fi

counter=1

# 20 NABs to be created
while [ $counter -le 20 ]; do
  object_name="anothertestbackup-${counter}"
  
  cat <<EOF | oc apply -f -
kind: NonAdminBackup
metadata:
  name: ${object_name}
  namespace: ${BACKUP_NAMESPACE}
spec:
  backupSpec:
    snapshotVolumes: true
    snapshotMoveData: true
    defaultVolumesToFsBackup: true
EOF

  echo "Created NonAdminBackup: ${object_name} in namespace ${BACKUP_NAMESPACE}"

  ((counter++))

  # Wait for 1 second before creating the next object
  sleep 1
done
```
- Observe estimatedQueuePositon within NAB Objects, can be done with:
```shell
$ while true; do oc get nonadminbackup -n "$BACKUP_NAMESPACE" -o jsonpath='{range .items[*]}{.status.queueInfo}{"\t"}{.metadata.name}{"\n"}{end}'|sort; sleep 1; done
```


api/v1alpha1/nonadminbackup_types.go
```diff
@@ -105,6 +105,16 @@ type NonAdminBackupStatus struct {
...
 	Conditions []metav1.Condition  `json:"conditions,omitempty"`
+
+	// +optional
+	QueueInfo *QueueInfo `json:"queueInfo,omitempty"`
+}
+
+// QueueInfo holds the queue position for a specific VeleroBackup.
+// It is used to determine how many backups are scheduled before the given VeleroBackup in the OADP namespace.
+// and counts only VeleroBackups that are still subject to be handled by OADP/Velero.
+type QueueInfo struct {
```
shubham-pampattiwar, line 116: Do we need a QueueInfo struct here ? Is the idea that we want to keep this flexible for future enhancements ? like actually giving the names of the backups that are ahead in the processing queue ?

config/crd/bases/oadp.openshift.io_nonadminbackups.yaml
```diff
@@ -619,6 +619,17 @@ spec:
                 - Created
                 - Deleting
                 type: string
+              queueInfo:
+                description: |-
```
shubham-pampattiwar, line 623: Please also create the corresponding OADP Operator PR so that we merged the CRD changes first and CI passess here.

internal/common/function/function.go
```diff
@@ -161,6 +161,69 @@ func GetVeleroBackupByLabel(ctx context.Context, clientInstance client.Client, n
...
+}
+
+// GetBackupQueueInfo determines the queue position of the specified VeleroBackup.
+// It calculates how many queued Backups exist in the namespace that were created before this one.
+func GetBackupQueueInfo(ctx context.Context, clientInstance client.Client, namespace string, targetBackup *velerov1.Backup) (nacv1alpha1.QueueInfo, error) {
+	var queueInfo nacv1alpha1.QueueInfo
+
+	// If the target backup has no valid CreationTimestamp, it means that it's not yet reconciled by OADP/Velero.
+	// In this case, we can't determine its queue position, so we return nil.
+	if targetBackup == nil || targetBackup.CreationTimestamp.IsZero() {
```
shubham-pampattiwar, line 196: AFAIK creationTS field gets populated as soon as the object create request is executed by Kube API server, it does not signify whether the controller of the object reconciled on it or not.

internal/common/function/function.go
```diff
@@ -161,6 +161,69 @@ func GetVeleroBackupByLabel(ctx context.Context, clientInstance client.Client, n
...
+	for i := range backupList.Items {
+		backup := &backupList.Items[i]
+
+		// Skip backups that have CompletionTimestamp set. This means that the Velero won't be further processing this backup.
+		if backup.Status.CompletionTimestamp != nil {
+			continue
+		}
+
+		// Count backups created earlier than the target backup
+		if backup.CreationTimestamp.Time.Before(targetTimestamp) {
```
shubham-pampattiwar, line 219: ++

internal/handler/velerobackup_queue_handler.go
```diff
@@ -0,0 +1,96 @@
...
+	logger := function.GetLogger(ctx, evt.ObjectNew, "VeleroBackupQueueHandler")
+
+	backups, err := function.GetActiveVeleroBackupsByLabel(ctx, h.Client, h.OADPNamespace, constant.ManagedByLabel, constant.ManagedByLabelValue)
+	if err != nil {
+		logger.Error(err, "Failed to get Velero Backups by label")
+		return
+	}
+
+	if backups == nil {
+		// That should't really be the case as our Update event was triggered by a Velero Backup
```
shubham-pampattiwar, line 60: NIT: shouldn't

internal/handler/velerobackup_queue_handler.go
```diff
@@ -0,0 +1,96 @@
...
+// Update event handler adds Velero Backup's NonAdminBackup to controller queue
+func (h VeleroBackupQueueHandler) Update(ctx context.Context, evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	// Only update to the first in the queue Velero Backup should trigger changes to the
+	// NonAdminBackup objects. Updates to the Velero Backup 2nd and 3rd does not lower the
+	// queue. This optimizes the number of times we need to update the NonAdminBackup objects
+	// and the number of Velero Backup objects we need to react on.
+
+	logger := function.GetLogger(ctx, evt.ObjectNew, "VeleroBackupQueueHandler")
+
+	backups, err := function.GetActiveVeleroBackupsByLabel(ctx, h.Client, h.OADPNamespace, constant.ManagedByLabel, constant.ManagedByLabelValue)
```
shubham-pampattiwar, line 53: just add a comment that we are trying to fetch VBs triggered by NAB and we want to optimize our reconcile cycles.

api/v1alpha1/nonadminbackup_types.go
mpryc: I was thinking more about stats like time to completion or size of backups ahead of current one, that's why created struct and not a single value to not limit ourselves. But can change to a single val.

shubham-pampattiwar: yeah makes sense, lets keep it is as struct.

internal/handler/velerobackup_queue_handler.go
mpryc: done

config/crd/bases/oadp.openshift.io_nonadminbackups.yaml
mpryc: https://github.com/openshift/oadp-operator/pull/1598

shubham-pampattiwar (approved):

config/crd/bases/oadp.openshift.io_nonadminbackups.yaml
```diff
@@ -619,6 +619,19 @@ spec:
...
+                  QueueInfo holds the queue position for a specific VeleroBackup.
+                  It is used to estimate how many backups are scheduled before the given VeleroBackup in the OADP namespace.
+                  This number is not guaranteed to be accurate, but it should be close. It's inaccurate for cases when
+                  Velero pod is not running or being restarted after Backup object were created.
+                  It counts only VeleroBackups that are still subject to be handled by OADP/Velero.
+                properties:
+                  estimatedQueuePosition:
+                    type: integer
+                required:
```
mateusoliveira43, line 632: why it is required?

mpryc: To be consistent across all NAB objects. Our controller does the creation and handling so it's up to us.

PR #121, opened 2024-11-29
fix: Import OADP label from OADP

mateusoliveira43:
## Why the changes were made

Use value from OADP instead of defining it.

Check CI logs.


mpryc (commented):
I like this change.

shubham-pampattiwar (approved):

kaovilai (approved):

PR #122, opened 2024-12-03
fix: Enforce NonAdminRestore spec field values

mateusoliveira43:
## Why the changes were made

Fix #37 

Blocked by https://github.com/openshift/oadp-operator/pull/1600

This PR adds DPA `spec.nonAdmin.enforceRestoreSpec` to `NonAdminRestoreReconciler` and adds validation for NonAdminRestore spec respecting enforcement.

> **Note:** admin user can not enforce falsy values.

## How to test the changes made

Read [design](https://github.com/mateusoliveira43/oadp-non-admin/blob/fix/enforce-restore-spec-field-values/docs/design/admin_control_over_spec.md).

Follow [install-from-source](https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#install-from-source) testing documentation, pointing this branch and my OADP fork branch [`fix/enforce-non-admin-restore-spec-field-values`](https://github.com/mateusoliveira43/oadp-operator/tree/fix/enforce-non-admin-restore-spec-field-values)

PR #124, opened 2024-12-04
Fix "cannot range over" local lint errors

mpryc:
With the #122 local error on the make lint
appeared. This is not visible in the CI, however local go version and lint version are the same, so may be associated with other dependent libraries.

Still this is an issue when submitting a PR from the system where this linting error is visible.

For error message and insights:
  https://github.com/migtools/oadp-non-admin/pull/122#issuecomment-2515304578

## Why the changes were made

To fix local `$ make lint` errors introduced with #122 

Ran the lint after change, no error appeared.

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -145,7 +145,7 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
 	// TODO nonAdminRestore.Spec.RestoreSpec.NamespaceMapping ?
 
 	enforcedSpec := reflect.ValueOf(enforcedRestoreSpec).Elem()
-	for index := range enforcedSpec.NumField() {
+	for index := 0; index < enforcedSpec.NumField(); index++ {
 		enforcedField := enforcedSpec.Field(index)
 		enforcedFieldName := enforcedSpec.Type().Field(index).Name
 		currentField := reflect.ValueOf(nonAdminRestore.Spec.RestoreSpec).Elem().FieldByName(enforcedFieldName)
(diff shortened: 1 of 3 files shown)


kaovilai:
Was it even working? I don't think golang allows range over int.

mpryc:
I think we may have CI github action lint br8ken or some wired casting which github lint doesn't understand. Wired is that @mateusoliveira43 can not reproduce this error, where I can on a fresh clone without go deps pre-cached.

mateusoliveira43:
https://stackoverflow.com/a/21951131

kaovilai (approved):

mateusoliveira43:
That is the problem

I am using
```go
$ ./bin/golangci-lint-v1.56.2 --version
golangci-lint has version v1.56.2 built with go1.22.6 from (unknown, mod sum: "h1:dgQzlWHgNbCqJjuxRJhFEnHDVrrjuTGQHJ3RIZMpp/o=") on (unknown)
```

kaovilai:
https://github.com/golangci/golangci-lint/blob/794a340e63eb3c76cd904e590b5909c6c7e132d4/pkg/golinters/typecheck.go#L10

kaovilai:
https://github.com/golangci/golangci-lint/issues/4459 says to use 1.56.0 built with go1.22

kaovilai:
Can we fix linter instead of working around linter?

kaovilai:
https://github.com/migtools/oadp-non-admin/blob/27679313ad6e7b4eaf53e59f3e0ae5a30947523a/go.mod#L86
is this being followed up later?

kaovilai:
I favor #126 let's close this one.

PR #126, opened 2024-12-05
Fix "cannot range over" local lint errors

mpryc:
Fix for the "cannot range over" that bumps the golang-ci version Makes simila range loop in the nab controller.

## Why the changes were made

To fix local $ make lint errors introduced with https://github.com/migtools/oadp-non-admin/pull/122

New version of golang-ci is fixing the issue on local ran.

ran the lint after change, no error appeared.

This PR is similar to #124, but uses different approach, so one or other should be merged.


mpryc:
@kaovilai FYI

mateusoliveira43 (changes requested):
need to change here https://github.com/migtools/oadp-non-admin/blob/2113181193f1b549482ce1d5a5b985b52e94d2d2/.golangci.yml#L1

kaovilai (approved):

kaovilai (approved):

kaovilai:
Why did I not have issue with 1.56.2 tho.. 🤷 

https://github.com/migtools/oadp-non-admin/pull/124#pullrequestreview-2479278049

mateusoliveira43 (approved):

mateusoliveira43:
@kaovilai https://github.com/migtools/oadp-non-admin/pull/124#issuecomment-2517847473

Michal's executable was built with go 1.21, mine and yours with 1.22

kaovilai:
Too bad.. I think you should avoid using go install for linters.. that would solve this issue . That meant this PR is no-op.

kaovilai:
perhaps the makefile should set GOTOOLCHAIN, and only then can the `make lint` be reliable :D

https://go.dev/doc/toolchain

PR #127, opened 2024-12-06
fix: CI flakes

mateusoliveira43:
## Why the changes were made

TODO

Check CI logs

### DIFF
--- a/internal/controller/nonadminbackup_controller.go
+++ b/internal/controller/nonadminbackup_controller.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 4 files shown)


internal/controller/nonadminbackup_controller.go
```diff
@@ -175,19 +173,10 @@ func (r *NonAdminBackupReconciler) setStatusAndConditionForDeletionAndCallDelete
 			return false, err
 		}
 		logger.V(1).Info("NonAdminBackup status marked for deletion")
-		requeueRequired = true // Requeue to ensure latest NAB object in the next reconciliation steps
 	} else {
 		logger.V(1).Info("NonAdminBackup status unchanged during deletion")
 	}
-	if nab.ObjectMeta.DeletionTimestamp.IsZero() {
-		logger.V(1).Info("Marking NonAdminBackup for deletion", constant.NameString, nab.Name)
-		if err := r.Delete(ctx, nab); err != nil {
```
mateusoliveira43, line 184: once this is called, update predicate is triggered. So, this should be last step to avoid running multiple reconciles

mpryc: But if you just return `false, nil` then reconcile will enter another step and will not exit as last step?

mpryc: I think we may need a 3rd state as it was in the very initial implementation which means "exit cleanly now" to avoid complexity moving parts to the steps where they don't belong. Moving this to the last step e.g. `removeNabFinalizerUponVeleroBackupDeletion` also doesn't makes sense to me as this step is actually to remove finalizer and not call delete on an object.

mateusoliveira43: what I did was add this as the last step of deletion, so `return false` will not do anything more

mpryc: How this works with the path where Deletion happens via `DeleteBackupRequest`

mateusoliveira43: it should be as it was, only change was the order of steps

mpryc:
@mateusoliveira43 is this flake still visible? I can see the log from the issue is from 27th of September. There were so many changes in the reconcile paths since then, that I wonder if it's even still reproducible?

mateusoliveira43:
@mpryc do not remember seeing it lately

But with the approach of removing `sleep` we can be safe they will not happen again. So I think is ok to close issue with the merge of this PR

mpryc:
@mateusoliveira43 right, I am just not comfortable with reworking reconcile to remove one sleep as this PR is much more then just one issue fix and I am still not sold on the rework here.

PR #128, opened 2024-12-06
Implementation for the restore queue

mpryc:
Adds the estimated queue to the NAB restore object

## Why the changes were made

To include estimated queue number for the restore object.

1. Create multiple NAR objects and observe if the queue is proper.
2. Run tests from the source `make simulation-test`
```shell
$ make simulation-test
[...]
Ran 19 of 19 Specs in 23.446 seconds
SUCCESS! -- 19 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestControllers (23.45s)
coverage: 64.6% of statements
```

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -314,6 +314,79 @@ func GetBackupQueueInfo(ctx context.Context, clientInstance client.Client, names
 	return queueInfo, nil
 }
 
+// GetActiveVeleroRestoresByLabel retrieves all VeleroRestore objects based on a specified label within a given namespace.
+// It returns a slice of VeleroRestore objects or nil if none are found.
+func GetActiveVeleroRestoresByLabel(ctx context.Context, clientInstance client.Client, namespace, labelKey, labelValue string) ([]velerov1.Restore, error) {
+	var veleroRestoreList velerov1.RestoreList
+	labelSelector := client.MatchingLabels{labelKey: labelValue}
+
+	if err := clientInstance.List(ctx, &veleroRestoreList, client.InNamespace(namespace), labelSelector); err != nil {
+		return nil, err
+	}
+
+	// Filter out restores with a CompletionTimestamp
+	var activeRestores []velerov1.Restore
+	for _, restore := range veleroRestoreList.Items {
+		if restore.Status.CompletionTimestamp == nil {
+			activeRestores = append(activeRestores, restore)
+		}
+	}
+
+	if len(activeRestores) == 0 {
+		return nil, nil
+	}
+
+	return activeRestores, nil
+}
+
+// GetRestoreQueueInfo determines the queue position of the specified VeleroRestore.
+// It calculates how many queued Restores exist in the namespace that were created before this one.
+func GetRestoreQueueInfo(ctx context.Context, clientInstance client.Client, namespace string, targetRestore *velerov1.Restore) (nacv1alpha1.QueueInfo, error) {
+	var queueInfo nacv1alpha1.QueueInfo
+
+	// If the target restore has no valid CreationTimestamp, it means that it's not yet reconciled by OADP/Velero.
+	// In this case, we can't determine its queue position, so we return nil.
+	if targetRestore == nil || targetRestore.CreationTimestamp.IsZero() {
+		return queueInfo, nil
+	}
+
+	// If the target restore has a CompletionTimestamp, it means that it's already served.
+	if targetRestore.Status.CompletionTimestamp != nil {
+		queueInfo.EstimatedQueuePosition = 0
+		return queueInfo, nil
+	}
+
+	// List all Restore objects in the namespace
+	var restoreList velerov1.RestoreList
+	if err := clientInstance.List(ctx, &restoreList, client.InNamespace(namespace)); err != nil {
+		return queueInfo, err
+	}
+
+	// Extract the target restore's creation timestamp
+	targetTimestamp := targetRestore.CreationTimestamp.Time
+
+	// The target restore is always in queue at least in the first position
+	// 0 is reserved for the restores that are already served.
+	queueInfo.EstimatedQueuePosition = 1
+
+	// Iterate through restores and calculate position
+	for i := range restoreList.Items {
+		restore := &restoreList.Items[i]
+
+		// Skip restores that have CompletionTimestamp set. This means that the Velero won't be further processing this restore.
+		if restore.Status.CompletionTimestamp != nil {
+			continue
+		}
+
+		// Count restores created earlier than the target restore
+		if restore.CreationTimestamp.Time.Before(targetTimestamp) {
+			queueInfo.EstimatedQueuePosition++
+		}
+	}
+
+	return queueInfo, nil
+}
+
 // GetVeleroDeleteBackupRequestByLabel retrieves a DeleteBackupRequest object based on a specified label within a given namespace.
 // It returns the DeleteBackupRequest only when exactly one object is found, throws an error if multiple backups are found,
 // or returns nil if no matches are found.
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -423,7 +423,7 @@ func GetVeleroRestoreByLabel(ctx context.Context, clientInstance client.Client, 
 	case 1:
 		return &veleroRestoreList.Items[0], nil
 	default:
-		return nil, fmt.Errorf("multiple Velero Restore objects found with label %s=%s in namespace '%s'", constant.NarOriginNACUUIDLabel, labelValue, namespace)
+		return nil, fmt.Errorf("multiple VeleroRestore objects found with label %s=%s in namespace '%s'", constant.NarOriginNACUUIDLabel, labelValue, namespace)
 	}
 }
 
--- a/internal/common/function/function_test.go
+++ b/internal/common/function/function_test.go
@@ -47,19 +47,20 @@ import (
 var _ = ginkgo.Describe("PLACEHOLDER", func() {})
 
 const (
-	testNonAdminBackupNamespace  = "non-admin-backup-namespace"
-	testNonAdminBackupName       = "non-admin-backup-name"
-	testNonAdminSecondBackupName = "non-admin-second-backup-name"
-	testNonAdminBackupUUID       = "12345678-1234-1234-1234-123456789abc"
-	testNonAdminRestoreName      = "non-admin-restore-name"
-	testNonAdminRestoreUUID      = "12345678-1234-1234-1234-123456789abc"
-	testNonAdminRestoreNamespace = "non-admin-restore-namespace"
-	defaultStr                   = "default"
-	invalidInputEmptyNamespace   = "Invalid input - empty namespace"
-	invalidInputEmptyNsErr       = "invalid input: namespace, labelKey, and labelValue must not be empty"
-	expectedIntZero              = 0
-	expectedIntOne               = 1
-	expectedIntTwo               = 2
+	testNonAdminBackupNamespace   = "non-admin-backup-namespace"
+	testNonAdminBackupName        = "non-admin-backup-name"
+	testNonAdminSecondBackupName  = "non-admin-second-backup-name"
+	testNonAdminBackupUUID        = "12345678-1234-1234-1234-123456789abc"
+	testNonAdminRestoreName       = "non-admin-restore-name"
+	testNonAdminRestoreUUID       = "12345678-1234-1234-1234-123456789abc"
+	testNonAdminRestoreNamespace  = "non-admin-restore-namespace"
+	testNonAdminSecondRestoreName = "non-admin-second-restore-name"
+	defaultStr                    = "default"
+	invalidInputEmptyNamespace    = "Invalid input - empty namespace"
+	invalidInputEmptyNsErr        = "invalid input: namespace, labelKey, and labelValue must not be empty"
+	expectedIntZero               = 0
+	expectedIntOne                = 1
+	expectedIntTwo                = 2
 )
 
 func TestGetNonAdminLabels(t *testing.T) {
@@ -1484,6 +1485,91 @@ func TestGetBackupQueueInfo(t *testing.T) {
 			client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(objects...).Build()
 
 			queueInfo, err := GetBackupQueueInfo(ctx, client, tt.namespace, tt.targetBackup)
+			assert.NoError(t, err)
+			assert.Equal(t, tt.expectedQueue, queueInfo.EstimatedQueuePosition)
+		})
+	}
+}
+
+func TestGetRestoreQueueInfo(t *testing.T) {
+	log := zap.New(zap.UseDevMode(true))
+	ctx := context.Background()
+	ctx = ctrl.LoggerInto(ctx, log)
+	scheme := runtime.NewScheme()
+
+	if err := velerov1.AddToScheme(scheme); err != nil {
+		t.Fatalf("Failed to register VeleroRestore type in TestGetRestoreQueueInfo: %v", err)
+	}
+
+	tests := []struct {
+		name          string
+		namespace     string
+		targetRestore *velerov1.Restore
+		mockRestores  []velerov1.Restore
+		expectedQueue int
+	}{
+		{
+			name:      "No restores in queue",
+			namespace: defaultStr,
+			targetRestore: &velerov1.Restore{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace:         defaultStr,
+					Name:              testNonAdminRestoreName,
+					CreationTimestamp: metav1.Time{Time: time.Now()},
+				},
+			},
+			mockRestores:  []velerov1.Restore{},
+			expectedQueue: expectedIntOne,
+		},
+		{
+			name:      "One restore ahead in queue",
+			namespace: defaultStr,
+			targetRestore: &velerov1.Restore{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace:         defaultStr,
+					Name:              testNonAdminSecondRestoreName,
+					CreationTimestamp: metav1.Time{Time: time.Now().Add(1 * time.Hour)},
+				},
+			},
+			mockRestores: []velerov1.Restore{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace:         defaultStr,
+						Name:              testNonAdminRestoreName,
+						CreationTimestamp: metav1.Time{Time: time.Now()},
+					},
+				},
+			},
+			expectedQueue: expectedIntTwo,
+		},
+		{
+			name:      "Target restore already completed",
+			namespace: defaultStr,
+			targetRestore: &velerov1.Restore{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace:         defaultStr,
+					Name:              testNonAdminRestoreName,
+					CreationTimestamp: metav1.Time{Time: time.Now()},
+				},
+				Status: velerov1.RestoreStatus{
+					CompletionTimestamp: &metav1.Time{Time: time.Now()},
+				},
+			},
+			mockRestores:  []velerov1.Restore{},
+			expectedQueue: expectedIntZero,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var objects []client.Object
+			for _, restore := range tt.mockRestores {
+				restoreCopy := restore
+				objects = append(objects, &restoreCopy)
+			}
+			client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(objects...).Build()
+
+			queueInfo, err := GetRestoreQueueInfo(ctx, client, tt.namespace, tt.targetRestore)
 			assert.NoError(t, err)
 			assert.Equal(t, tt.expectedQueue, queueInfo.EstimatedQueuePosition)
 		})
--- a/internal/common/function/function_test.go
+++ b/internal/common/function/function_test.go
@@ -51,7 +51,12 @@ const (
 	testNonAdminBackupName       = "non-admin-backup-name"
 	testNonAdminSecondBackupName = "non-admin-second-backup-name"
 	testNonAdminBackupUUID       = "12345678-1234-1234-1234-123456789abc"
+	testNonAdminRestoreName      = "non-admin-restore-name"
+	testNonAdminRestoreUUID      = "12345678-1234-1234-1234-123456789abc"
+	testNonAdminRestoreNamespace = "non-admin-restore-namespace"
 	defaultStr                   = "default"
+	invalidInputEmptyNamespace   = "Invalid input - empty namespace"
+	invalidInputEmptyNsErr       = "invalid input: namespace, labelKey, and labelValue must not be empty"
 	expectedIntZero              = 0
 	expectedIntOne               = 1
 	expectedIntTwo               = 2
@@ -826,12 +831,12 @@ func TestGetVeleroBackupByLabel(t *testing.T) {
 			expectedError: errors.New("multiple VeleroBackup objects found with label openshift.io/oadp-nab-origin-nacuuid=test-app in namespace 'default'"),
 		},
 		{
-			name:          "Invalid input - empty namespace",
-			namespace:     "",
+			name:          invalidInputEmptyNamespace,
+			namespace:     constant.EmptyString,
 			labelValue:    testAppStr,
 			mockBackups:   []velerov1.Backup{},
 			expected:      nil,
-			expectedError: errors.New("invalid input: namespace, labelKey, and labelValue must not be empty"),
+			expectedError: errors.New(invalidInputEmptyNsErr),
 		},
 	}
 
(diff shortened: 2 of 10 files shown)


mpryc:
@mateusoliveira43 note the file where I've moved the common types. Let me know if that is fine, currently we were mixing backup types inside restore object: https://github.com/migtools/oadp-non-admin/pull/128/files#diff-bce3cc24a9a3be49f46a28f3a11baff2a1f07f5e735a1dbc683e535e1aaf625d

mateusoliveira43:
@mpryc it is fine

I would change name to something like `common.go` to avoid any possible conflict with kubebuilder in the future

api/v1alpha1/nonadmin_types.go
```diff
@@ -0,0 +1,41 @@
...
+	NonAdminPhaseNew NonAdminPhase = "New"
+	// NonAdminPhaseBackingOff - Velero object was not created due to NonAdmin object error (configuration or similar)
+	NonAdminPhaseBackingOff NonAdminPhase = "BackingOff"
+	// NonAdminPhaseCreated - Velero object was created. The Phase will not have additional information about it.
+	NonAdminPhaseCreated NonAdminPhase = "Created"
+	// NonAdminPhaseDeleting - Velero object is pending deletion. The Phase will not have additional information about it.
+	NonAdminPhaseDeleting NonAdminPhase = "Deleting"
+)
+
+// QueueInfo represents the position of a specific operation request (Velero Backup or Restore) in the queue.
```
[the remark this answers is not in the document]
mpryc: I wanted to make it common - do you think it's worth duplicating the structure if it's exactly the same for all the objects?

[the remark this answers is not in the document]
mateusoliveira43: not the struct, just the comment

adding here for example https://github.com/mpryc/oadp-non-admin/blob/queue-restore/api/v1alpha1/nonadminbackup_types.go#L86

mpryc:
Blocked by: https://github.com/openshift/oadp-operator/pull/1607

mateusoliveira43:
@mpryc avoid using `make test` and use `make simulation-test`, as described in our documentation https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#code-quality-and-standardization

internal/handler/velerorestore_queue_handler.go
mateusoliveira43, line 1: I do not think we need new predicate and handler for queue

doing these changes on the current ones, does not give the same result?
```diff
diff --git a/internal/handler/velerorestore_handler.go b/internal/handler/velerorestore_handler.go
index 515de66..ff2b23c 100644
--- a/internal/handler/velerorestore_handler.go
+++ b/internal/handler/velerorestore_handler.go
@@ -21,6 +21,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -29,7 +30,10 @@ import (
 )
 
 // VeleroRestoreHandler contains event handlers for Velero Restore objects
-type VeleroRestoreHandler struct{}
+type VeleroRestoreHandler struct {
+	Client        client.Client
+	OADPNamespace string
+}
 
 // Create event handler
 func (VeleroRestoreHandler) Create(_ context.Context, _ event.CreateEvent, _ workqueue.RateLimitingInterface) {
@@ -37,18 +41,42 @@ func (VeleroRestoreHandler) Create(_ context.Context, _ event.CreateEvent, _ wor
 }
 
 // Update event handler adds Velero Restore's NonAdminRestore to controller queue
-func (VeleroRestoreHandler) Update(ctx context.Context, evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+func (h VeleroRestoreHandler) Update(ctx context.Context, evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
 	logger := function.GetLogger(ctx, evt.ObjectNew, "VeleroRestoreHandler")
 
 	annotations := evt.ObjectNew.GetAnnotations()
 	nonAdminRestoreName := annotations[constant.NarOriginNameAnnotation]
 	nonAdminRestoreNamespace := annotations[constant.NarOriginNamespaceAnnotation]
 
-	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
-		Name:      nonAdminRestoreName,
-		Namespace: nonAdminRestoreNamespace,
-	}})
-	logger.V(1).Info("Handled Update event")
+	if function.CheckVeleroRestoreMetadata(evt.ObjectNew) {
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+			Name:      nonAdminRestoreName,
+			Namespace: nonAdminRestoreNamespace,
+		}})
+		logger.V(1).Info("Handled Update event")
+	}
+
+	restores, err := function.GetActiveVeleroRestoresByLabel(ctx, h.Client, h.OADPNamespace)
+	if err != nil {
+		logger.Error(err, "Failed to get Velero Restores by label")
+		return
+	}
+
+	if restores != nil {
+		for _, restore := range restores {
+			annotations := restore.GetAnnotations()
+			originName := annotations[constant.NarOriginNameAnnotation]
+			originNamespace := annotations[constant.NarOriginNamespaceAnnotation]
+
+			if originName != nonAdminRestoreName || originNamespace != nonAdminRestoreNamespace {
+				logger.V(1).Info("Processing Queue update for the NonAdmin Restore referenced by Velero Restore", "Name", restore.Name, constant.NamespaceString, restore.Namespace, "CreatedAt", restore.CreationTimestamp)
+				q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+					Name:      originName,
+					Namespace: originNamespace,
+				}})
+			}
+		}
+	}
 }
 
 // Delete event handler
diff --git a/internal/predicate/velerorestore_predicate.go b/internal/predicate/velerorestore_predicate.go
index 8dfdae8..e63e75b 100644
--- a/internal/predicate/velerorestore_predicate.go
+++ b/internal/predicate/velerorestore_predicate.go
@@ -19,6 +19,7 @@ package predicate
 import (
 	"context"
 
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	"github.com/migtools/oadp-non-admin/internal/common/function"
@@ -40,6 +41,12 @@ func (p VeleroRestorePredicate) Update(ctx context.Context, evt event.UpdateEven
 			logger.V(1).Info("Accepted Update event")
 			return true
 		}
+		newRestore, _ := evt.ObjectNew.(*velerov1.Restore)
+		oldRestore, _ := evt.ObjectOld.(*velerov1.Restore)
+		if oldRestore.Status.CompletionTimestamp == nil && newRestore.Status.CompletionTimestamp != nil {
+			logger.V(1).Info("Accepted Update event: Restore completion timestamp")
+			return true
+		}
 	}
 
 	logger.V(1).Info("Rejected Update event")

```

internal/common/function/function.go
```diff
@@ -314,6 +314,79 @@ func GetBackupQueueInfo(ctx context.Context, clientInstance client.Client, names
 	return queueInfo, nil
 }
 
+// GetActiveVeleroRestoresByLabel retrieves all VeleroRestore objects based on a specified label within a given namespace.
+// It returns a slice of VeleroRestore objects or nil if none are found.
+func GetActiveVeleroRestoresByLabel(ctx context.Context, clientInstance client.Client, namespace, labelKey, labelValue string) ([]velerov1.Restore, error) {
+	var veleroRestoreList velerov1.RestoreList
+	labelSelector := client.MatchingLabels{labelKey: labelValue}
+
+	if err := clientInstance.List(ctx, &veleroRestoreList, client.InNamespace(namespace), labelSelector); err != nil {
```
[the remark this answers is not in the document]
mpryc: I want all the NAC owned restores that does not have `restore.Status.CompletionTimestamp`, which means they are still subject to be handled. We are making queue number based on the list of the objects that are still subject to some work by Velero, the ones that are completed (has the CompletionTimestamp) are not valid anymore.

Will change the implementation to use the `ListObjectsByLabel`.

internal/handler/velerorestore_queue_handler.go
mpryc: Here I think we should stick with handler/predicate approach for two reasons:

1. Be similar with how backup works: 
- https://github.com/migtools/oadp-non-admin/blob/56afada606fe9e8d6ec96d1a1c0228c6d9d9c6fc/internal/predicate/velerobackup_queue_predicate.go#L37-L60
- https://github.com/migtools/oadp-non-admin/blob/56afada606fe9e8d6ec96d1a1c0228c6d9d9c6fc/internal/handler/velerobackup_queue_handler.go#L45

2. In k8s the predicate comes *before* handler, so we won't be invoking reconcile for cases where we can save some calls, in the above we are saving when the `CompletionTimestamp` is set from nil to something, meaning the object is "done". Only those are interested for us in regards how queue works. Those objects may have came to queue before ones that needs updating queue info. Anything that happened later is not interested as those are put later in the queue, meaning those will not affect any other objects that are in front. This allows to save quite a bit of calls on the predicate level. Secondly you are removing annotations check from the handler, which makes reconcile responsible for it. Reconcile comes after handler, which is also making more calls instead of reducing them.

internal/common/function/function.go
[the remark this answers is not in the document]
mpryc: Ah I remember why I could not use `ListObjectsByLabel` It's because velero `RestoreList`:

```
cannot use veleroRestoreList (variable of type "github.com/vmware-tanzu/velero/pkg/apis/velero/v1".RestoreList) as client.ObjectList value in argument to
ListObjectsByLabel: "github.com/vmware-tanzu/velero/pkg/apis/velero/v1".RestoreList does not implement client.ObjectList (method DeepCopyObject has pointer receiver)
```

So I am unable to change this as you requested.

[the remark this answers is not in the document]
mpryc: Should this also be changed? I don't want this to be any different? This PR/separate ?

https://github.com/migtools/oadp-non-admin/blob/56afada606fe9e8d6ec96d1a1c0228c6d9d9c6fc/internal/common/function/function.go#L246-L267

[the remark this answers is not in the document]
mateusoliveira43: I am ok doing change in this PR as well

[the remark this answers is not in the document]
mpryc: I will leave as it is in this PR if you agree - changed the backup function and now the tests are failing showing always queue 0. I don't want to spent time in this PR to debug why. Current implementation works as expected.

internal/handler/velerorestore_queue_handler.go
mateusoliveira43: I made some tests out of a cluster, and I am afraid we need to merge them together

If queue predicate is triggered, both handlers are triggered, right?

If a admin backup triggers this, velero backup predicate will try to add NonAdminBackup with empty name and namespace to the queue, right? and this creates and endless error

mpryc: I don't follow this use case?

The following handlers are triggered after predicates:
https://github.com/migtools/oadp-non-admin/blob/56afada606fe9e8d6ec96d1a1c0228c6d9d9c6fc/internal/controller/nonadminbackup_controller.go#L699-L700

So both handlers are triggered based on the composite predicate, each handler does different job.

This one only for the object which actually triggered the event:
   https://github.com/migtools/oadp-non-admin/blob/master/internal/handler/velerobackup_handler.go#L41-L53

And this one for all other objects to update statuses:
   https://github.com/migtools/oadp-non-admin/blob/master/internal/handler/velerobackup_handler.go#L41-L53
   
   See the namespace thing.

mateusoliveira43: if a backup not owned by NAC updates to completed, does that trigger the queue predicate? yes, right?

then, both handlers are called, and `velerobackup_handler.go` will try to add NonAdminBackup with empty name and namespace to the queue, right? and this creates and endless error

mpryc: @mateusoliveira43 I don't observe that.

See the controller logs from the start when I do add not owned by NAC:

```shell
2024-12-12T09:49:19+01:00	INFO	setup	starting manager
2024-12-12T09:49:19+01:00	INFO	controller-runtime.metrics	Starting metrics server
2024-12-12T09:49:19+01:00	INFO	controller-runtime.metrics	Serving metrics server	{"bindAddress": ":8080", "secure": false}
2024-12-12T09:49:19+01:00	INFO	starting server	{"kind": "health probe", "addr": "[::]:8081"}
2024-12-12T09:49:19+01:00	INFO	Starting EventSource	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "source": "kind source: *v1alpha1.NonAdminBackup"}
2024-12-12T09:49:19+01:00	INFO	Starting EventSource	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "source": "kind source: *v1.Backup"}
2024-12-12T09:49:19+01:00	INFO	Starting EventSource	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "source": "kind source: *v1.Backup"}
2024-12-12T09:49:19+01:00	INFO	Starting Controller	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup"}
2024-12-12T09:49:19+01:00	INFO	Starting EventSource	{"controller": "nonadminrestore", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminRestore", "source": "kind source: *v1alpha1.NonAdminRestore"}
2024-12-12T09:49:19+01:00	INFO	Starting EventSource	{"controller": "nonadminrestore", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminRestore", "source": "kind source: *v1.Restore"}
2024-12-12T09:49:19+01:00	INFO	Starting Controller	{"controller": "nonadminrestore", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminRestore"}
2024-12-12T09:49:20+01:00	INFO	Starting workers	{"controller": "nonadminrestore", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminRestore", "worker count": 1}
2024-12-12T09:49:20+01:00	INFO	Starting workers	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "worker count": 1}
2024-12-12T09:49:43+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:43+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:43+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:43+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:45+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:45+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:45+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:45+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:49+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:49+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:49+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:49+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Rejected Backup Update event: no changes to the CompletionTimestamp in the VeleroBackup object	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Rejected Backup Update event	{"VeleroBackupPredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Accepted Backup Update event: new completion timestamp	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Handled Update event	{"VeleroBackupHandler": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	Accepted Backup Update event: new completion timestamp	{"VeleroBackupQueuePredicate": {"name":"create-backup-not-owned","namespace":"openshift-adp"}}
2024-12-12T09:49:50+01:00	DEBUG	No pending velero backups found in namespace.	{"VeleroBackupQueueHandler": {"name":"create-backup-not-owned","namespace":"openshift-adp"}, "Namespace": "openshift-adp"}
2024-12-12T09:49:50+01:00	DEBUG	NonAdminBackup Reconcile start	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "NonAdminBackup": {"name":""}, "namespace": "", "name": "", "reconcileID": "7b7df2f5-9e9b-4bae-8f71-6894198946eb"}
2024-12-12T09:49:50+01:00	DEBUG	NonAdminBackup.oadp.openshift.io "" not found	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "NonAdminBackup": {"name":""}, "namespace": "", "name": "", "reconcileID": "7b7df2f5-9e9b-4bae-8f71-6894198946eb"}
```

As you can see this ends cleanly. Still can't reproduce endless reconcile loop.

mateusoliveira43: yeah, the error was on my test env only

But we should not do this `2024-12-12T09:49:50+01:00	DEBUG	NonAdminBackup.oadp.openshift.io "" not found	{"controller": "nonadminbackup", "controllerGroup": "oadp.openshift.io", "controllerKind": "NonAdminBackup", "NonAdminBackup": {"name":""}, "namespace": "", "name": "", "reconcileID": "7b7df2f5-9e9b-4bae-8f71-6894198946eb"}`

mpryc: That's just the debug info from reconcile letting know we are actually skipping reconcile. I am not sure if there is some improvement to be done here to move that part to the handler but definatelly different PR then this.

mpryc: I really do prefer to have working and similar implementation on both parts. With test day proven this works I don't want to modify too much in this area at the moment as I want to focus on #36 and then improve if we find issues. This for me is not an issue, just small improvement in the implementation. Going off the reconcile is fine imo at this moment.

mpryc: A bit more why I think it's really not that important at the moment to make this effort:
 1. We already have tested and working implementation, possibly there are areas to improve, but we need to focus on other parts (sync controller, nabsl)
 2. The current flow is pretty clean we have
     a) predicates for each type of interesting event (first defence/filter against unnecessary reconciles)
     b) separate handlers for velero events that works with:
         Current NAB object: https://github.com/migtools/oadp-non-admin/blob/56afada606fe9e8d6ec96d1a1c0228c6d9d9c6fc/internal/handler/velerobackup_handler.go#L48
         Other interested NAB objects: https://github.com/migtools/oadp-non-admin/blob/master/internal/handler/velerobackup_queue_handler.go#L74-L81

The two above are not causing the same NAB to be reconciled twice.

If we want to modify and check if velero object event is not triggering the reconcile on non-existing NAB we would need to move the check if NAB really exists to the handler, which is not the way it should work. Reconcile function is the proper place for this check, as it centralizes the logic, ensuring there is only one point where the NAB's existence is verified, rather than duplicating this check in the handler.

Aren't we rewriting yet one more time reconcile?

shubham-pampattiwar: @mateusoliveira43 @mpryc Lets keep the restore queue work similar to what we had in backup queue. If there are cases that are not being covered with the current approach, lets create an issue for those and then the issues can be fixed in follow on PRs.

internal/controller/nonadminrestore_controller.go
```diff
@@ -376,7 +376,18 @@ func (r *NonAdminRestoreReconciler) createVeleroRestore(ctx context.Context, log
 		logger.Info("VeleroRestore successfully created")
 	}
 
-	// TODO(migi): do we need estimatedQueuePosition in VeleroRestoreStatus?
+	updatedQueueInfo := false
+
+	// Determine how many Backups are scheduled before the given VeleroRestore in the OADP namespace.
```
mateusoliveira43, line 381: typo `Backups`

mpryc: Fixed

internal/controller/nonadminrestore_controller_test.go
```diff
@@ -357,6 +364,9 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminRestore Controller"
 						LastTransitionTime: metav1.NewTime(time.Now()),
 					},
 				},
+				QueueInfo: &nacv1alpha1.QueueInfo{
```
mateusoliveira43, line 367: I do not know what happens if we tell Velero to restore a backup in progress. I suspect it fails 

To make this test scenario more real, would change `VeleroBackup.Status` to phase completed, have CompletionTimestamp and `QueueInfo.EstimatedQueuePosition` to be zero

internal/common/function/function_test.go
mateusoliveira43, line 1: there is no test for `GetRestoreQueueInfo` function, right? would be interesting to add it?

internal/controller/nonadminrestore_controller_test.go
mpryc: Will change this, it's just a test and will be more aligned with world case scenario.

That said, if a user attempts to restore an InProgress backup, it should still be added to the queue. When the time comes to process that InProgress backup, it will simply fail immediately, effectively reducing the queue size at that point

internal/controller/nonadminrestore_controller_test.go
```diff
@@ -100,6 +100,13 @@ func checkTestNonAdminRestoreStatus(nonAdminRestore *nacv1alpha1.NonAdminRestore
 			return fmt.Errorf("NonAdminRestore Status Conditions [%v] Message %v does not contain expected message %v", index, nonAdminRestore.Status.Conditions[index].Message, expectedStatus.Conditions[index].Message)
 		}
 	}
+
```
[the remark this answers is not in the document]
mpryc: Fixed

internal/common/function/function_test.go
mpryc: Added

mpryc:
@mateusoliveira43 @shubham-pampattiwar rebased and addressed comments in the 4th commit to this PR, so it's easier to review latest changes.

internal/controller/nonadminrestore_controller_test.go
```diff
@@ -388,7 +418,7 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminRestore Controller"
 				},
 			},
 		}),
-		ginkgo.Entry("Should update NonAdminRestore until it invalidates and then delete it", nonAdminRestoreFullReconcileScenario{
+		ginkgo.Entry("Should not include queueInfo for the given backup as NonAdminBackup is not ready to be restored (BackingOff)", nonAdminRestoreFullReconcileScenario{
```
mateusoliveira43, line 421: would not update test name

the test scenario did not suffer great changes

This info can be added as a comment in line 449

internal/controller/nonadminrestore_controller_test.go
mateusoliveira43: But this can not happen 
https://github.com/migtools/oadp-non-admin/blob/9ba3da37d941fa92da9c80294d505d90cc8c01b1/internal/common/function/function.go#L280-L284

internal/controller/nonadminrestore_controller_test.go
mpryc: I made it more precise of what it does, so why not have it better if I've changed status check that affected this run?

internal/controller/nonadminrestore_controller_test.go
mpryc: I understand, but the purpose of the test is to ensure that any future changes to the code don’t alter this behavior, correct? You’re suggesting modifying the test scenario to match the current implementation, but I believe we should take a more black-box approach to verify that the implementation behaves as expected.

mateusoliveira43 (approved):
/lgtm

PR #133, opened 2024-12-12
fix: test improvement

### LINKED ISSUE #93: CI flake in integration test
Saw a CI flake with integration tests

The Reconcile of NAB took more then one second and it broke the test

Example https://github.com/migtools/oadp-non-admin/actions/runs/11070490617/job/30760280482?pr=91#step:5:95

## Possible fixes

Increase timeout
https://github.com/migtools/oadp-non-admin/blob/282620c4f6055e696d5d81df1af124bfd5481b6c/internal/controller/nonadminbackup_controller_test.go#L505

Use `Eventually` approach
https://github.com/migtools/oadp-non-admin/blob/282620c4f6055e696d5d81df1af124bfd5481b6c/internal/controller/nonadminbackup_controller_test.go#L541

mateusoliveira43:
## Why the changes were made

Fix #93 
Remove `Sleep` calls in test

Use cache in manager to avoid test overlapping

Check if objects were deleted

Check if only the expected number of errors were logged 

Check CI logs.

### DIFF
--- a/internal/controller/nonadminbackup_controller_test.go
+++ b/internal/controller/nonadminbackup_controller_test.go
@@ -30,9 +30,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


kaovilai (changes requested):
unit tests failing

Issue #38, opened 2024-04-09
Add Non-Admin Backup Sync controller


shubham-pampattiwar:
For creation of Non-Admin restore CR, the non-admin user will refer the non-admin backup instance. Most times in disaster recovery scenarios the NS will cease to exist and with it the Non-Admin Backup CRs too. Therefore, it is desirable that once the application NS exists and it does not have the relevant user's non-admin backup CRs, then there should be a mechanism to re-create those in the user NS, so that the non-admin user can view and use them for creation of non-admin restore CR and trigger the restore operation. Ensuring the appropriate sync of Backups would be the primary responsibility of Non-Admin Backup Sync controller.

Another point of note is that, this work would put the infrastructure in place for enabling cross cluster non-admin backup/restore operations in the future.

weshayutin:
Only terminal state... completed or failed state crs

mateusoliveira43:
Example implementation https://github.com/vmware-tanzu/velero/blob/bc29471ed6e94eda2129a7a070080b4f11b420de/pkg/controller/backup_sync_controller.go

should it also have a user time input for how frequent sync?

mpryc:
1. Find all Velero Backups in the namespace where the controller resides
2. Check if those were created by NAC controller (labels)
3. Check if corresponding NABs are in the namespaces
4. If there is a namespace without corresponding NAB, create NAB, but with a state that won't reconcile main NAC reconcile loop that creates another Velero Backup

The sync should work on:
 - any new namespace creation event on the cluster
 - time based ?

mateusoliveira43:
just namespace creation event seems problematic 

Example: admin user configures new cluster and then installs OADP. All non admin namespace already exist once NAC is live, then sync will not happen, right?

do we also need design for this?

mpryc:
I think the following then would be ok:
 - on any namespace creation event
 - on start of the sync controller (regardless of the namespace creation)

shawn-hurley:
What happens when you have a create event that you don't get before an update occurs? You will end up with only an "update event" in the queue

mateusoliveira43:
my current idea is to use NonAdminBSL controller for this

Velero has a separate controller just for this, but it is anti pattern to have multiple controllers watching the same resource

for this, we would have DPA field for sync frequency and status field in NonAdminBSL like `lastSyncTime`

[...]

PR #138, opened 2025-01-09
feat: NonAdminBackup sync controller

mateusoliveira43:
## Why the changes were made

Fix https://github.com/migtools/oadp-non-admin/issues/38

~Blocked by https://github.com/migtools/oadp-non-admin/pull/139~

~Blocked by https://github.com/migtools/oadp-non-admin/issues/116~

~Blocked by https://github.com/migtools/oadp-non-admin/pull/152~

~Blocked by https://github.com/openshift/oadp-operator/pull/1624~

Once NAC Pod is created, NonAdminBackup Synchronizer controller will start, and will run periodically. Its frequency run is defined by admin user in DPA. By default 2 minutes. 

NonAdminBackup Synchronizer controller will:
- fetch all BSLs in OADP namespace, filtering those that are default or are NAC owned and its related NABSL exists
- fetch all NAC owned Backups in OADP namespace, filtering those that are in completed state and use one of the filtered BSLs (from previous step)
- check if related NonAdminBackup exists and its namespace exists, for each filtered Backup (from previous step)
- non-existent NonAdminBackups (but that its origin namespace exist), will be recreated 

Follow [install-from-source](https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#install-from-source) testing documentation, pointing this branch and my OADP fork branch [`feat/nac-nab-sync-controller`](https://github.com/mateusoliveira43/oadp-operator/tree/feat/nac-nab-sync-controller)

Create NonAdminBackups, using NABSL and default BSL from OADP namespace. Delete them from cluster (but not from object storage). They should be recreated after sync period is reached.


internal/controller/nonadminbackup_controller.go
```diff
@@ -89,6 +89,8 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	_, syncBackup := nab.Labels["openshift.io/oadp-nab-synced-from-nacuuid"]
```
mateusoliveira43, line 92: this was my idea to differ NABs created by user and by sync controller: adding a label to object

what you think about this approach?

internal/controller/sync_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	backupStorageLocation, err := function.GetVeleroBackupStorageLocationByLabel(
+		ctx, r.Client, r.OADPNamespace,
+		nonAdminBackupStorageLocation.Status.VeleroBackupStorageLocation.NACUUID,
+	)
+	if err != nil {
+		logger.Error(err, "Failed to get VeleroBackupStorageLocation")
+		return ctrl.Result{}, err
+	}
+	if backupStorageLocation == nil {
```
mateusoliveira43, line 79: one of the reasons I think sync controller should NOT be a separate controller. This check would not be necessary if it was part of NonAdminBackupStorageLocation controller

internal/controller/sync_controller.go
```diff
@@ -0,0 +1,237 @@
...
+		err = fmt.Errorf("VeleroBackupStorageLocation not yet created")
+		logger.Error(err, "VeleroBackupStorageLocation not yet created")
+		return ctrl.Result{}, err
+	}
+
+	// // OPTION 1: copy how Velero does
+	// // PROBLEM: NAC Pod will not have Velero Pod files
```
mateusoliveira43, line 86: We would need to mount secret and plugins to NAC Pod to allow this approach, right?

I do not know how much work would that be

internal/controller/sync_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	// 		},
+	// 	}
+	// 	nab.Spec.BackupSpec.StorageLocation = nonAdminBackupStorageLocation.Name
+	// 	r.Create(ctx, nab)
+	// }
+
+	// ------------------------------------------------------------------------
+
+	// OPTION 2: compare BSLs specs
+	// PROBLEM: if user deletes NABSL, and them recreates it, velero backup will point to a BSL that does not exist
```
mateusoliveira43, line 169: I would try this approach, but it is not perfect

We could document that there are limitations, if there no other options :(

internal/controller/sync_controller_test.go
```diff
@@ -0,0 +1,171 @@
...
+							Name: "cloud-credentials",
+						},
+						Key: "cloud",
+					},
+				},
+			},
+		}),
+		// ginkgo.Entry("Should check that NonAdminBackupStorageLocation update changes next sync time, then delete NonAdminBackupStorageLocation", nonAdminBackupStorageLocationFullReconcileScenario{}),
+		// ginkgo.Entry("Should sync only finished NonAdminBackups, then delete NonAdminBackupStorageLocation", nonAdminBackupStorageLocationFullReconcileScenario{}),
+		// ginkgo.Entry("Should sync only related NonAdminBackupStorageLocation NonAdminBackups, then delete NonAdminBackupStorageLocation", nonAdminBackupStorageLocationFullReconcileScenario{}),
```
mateusoliveira43, line 169: test some cross cluster scenarios as well

internal/controller/nonadminbackup_controller.go
mpryc: That sounds like a proper design for that. Would use something from `https://github.com/openshift/oadp-operator/blob/master/api/v1alpha1/oadp_types.go#L35` and we are using it in few places already in the constant.go

Possibly adding yet one label, but that would mean the object is managed by entirely and I think that may be not true, as once it's synced it's no longer managed by sync controller it's the user responsibility to do something with it like deletion:
```
app.kubernetes.io/managed-by: self-service-sync-controller
```

internal/controller/sync_controller.go
mpryc: I think the sync controller should restore the NAB objects based on the Backup objects that are found in the OADP namespace (by default `openshift-adp`). I am not sure if the sync controller should really go into s3 BSLs itself?

I do not know why there is anything with secrets to be done, or I am missing something.

Once the BSL is in the OADP namespace the secret is also in that namespace and corresponds to the BSL and not the NABSL.

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -56,10 +59,13 @@ type NonAdminBackupStorageLocationReconciler struct {
 	Scheme          *runtime.Scheme
 	EnforcedBslSpec *velerov1.BackupStorageLocationSpec
 	OADPNamespace   string
+	SyncPeriod      time.Duration
```
mateusoliveira43, line 62: TODO user input

internal/controller/sync_controller.go
shubham-pampattiwar: NaSYNC : openshift-adp <-> non-admin ns
Velero Sync: s3 <-> openshift-adp

cmd/main.go
```diff
@@ -169,6 +170,8 @@ func main() {
 		Client:        mgr.GetClient(),
 		Scheme:        mgr.GetScheme(),
 		OADPNamespace: oadpNamespace,
+		// TODO user input
+		SyncPeriod: 5 * time.Minute,
```
shubham-pampattiwar, line 174: yeah correct, this would be a user input but we would need a default value.
Have you given any thoughts on how to accept this from the user ? I can think of a new possibilities:
- From `DPA.spec.nonAdmin` struct.
- Create a struct under `DPA.spec.nonAdmin.Args` and accept the value there.
- Introduce a configmap for nacusers/admin users for specifying such things.
- All of the above are extensible enough for similar user inputs in the future NAC world if needed.

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -123,7 +129,71 @@ func (r *NonAdminBackupStorageLocationReconciler) Reconcile(ctx context.Context,
 		}
 	}
 
-	logger.V(1).Info("NonAdminBackup Reconcile exit")
+	if previousBackupSyncRun == nil || time.Now().After(previousBackupSyncRun.Add(r.SyncPeriod)) {
```
shubham-pampattiwar, line 132: Why not introduce a separate controller to sync the backups ? 
Also, if not a new controller then would it make sense to add this in NAB controller instead of NABSL ?

cmd/main.go
kaovilai: The minimum should be no less than velero's sync controller IMO.

internal/controller/nonadminbackupsynchronizer_controller.go
```diff
@@ -0,0 +1,168 @@
...
+		return ctrl.Result{RequeueAfter: r.SyncPeriod}, nil
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NonAdminBackupSynchronizerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&velerov1.BackupStorageLocation{}).
+		// WithEventFilter(predicate.GarbageCollectorPredicate{}).
```
mateusoliveira43, line 166: discuss what events to watch

internal/controller/nonadminbackupsynchronizer_controller.go
```diff
@@ -0,0 +1,168 @@
...
+		veleroBackupList := &velerov1.BackupList{}
+		if err := r.List(ctx, veleroBackupList, client.InNamespace(r.OADPNamespace), labelSelector); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		var backupsToSync []velerov1.Backup
+		var possibleBackupsToSync []velerov1.Backup
+		for _, backup := range veleroBackupList.Items {
+			if backup.Status.CompletionTimestamp != nil &&
+				slices.Contains(watchedBackupStorageLocations, backup.Spec.StorageLocation) {
```
mateusoliveira43, line 99: check non admin annotations

cmd/main.go
mateusoliveira43: it will be DPA field under nonAdmin 

I would not validate if the value is greater or equal to velero's value, do not know if is simple to be done

internal/controller/nonadminbackupstoragelocation_controller.go
mateusoliveira43: was to make logic simpler

with decision to use default BSLs from OADP namespace, its not feasible anymore

internal/controller/nonadminbackupsynchronizer_controller.go
mateusoliveira43, line 1: write tests

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: I think the default could be minute, similar to Velero Sync ?

https://github.com/vmware-tanzu/velero/blob/6ac38cde8556f0fb338f1b00cd64746790b6c73c/pkg/controller/backup_sync_controller.go#L51

mpryc: Would it be useful to use sync period from velero sync here? So do not have separate sync for each, just one config that will be for both (existing velero sync one)

shubham-pampattiwar: yes that would make sense. Import the default value also check for velero server arg if they are set by the admin.
Then pass it on to NAC controller. We can do this via env var on NAC controller. WDYT ? @mateusoliveira43 @mpryc

mpryc: I like this idea if it's easy to implement.

cmd/main.go
kaovilai: Velero value isn't customizable AFAICT so if you want to, it can be a hardcoded copy.

shubham-pampattiwar: @kaovilai https://github.com/vmware-tanzu/velero/blob/6ac38cde8556f0fb338f1b00cd64746790b6c73c/pkg/cmd/server/config/config.go#L233
Is this a different sync ?

kaovilai: checking.

kaovilai: looks like that will eventually set [here](https://github.com/vmware-tanzu/velero/blob/6ac38cde8556f0fb338f1b00cd64746790b6c73c/pkg/controller/backup_sync_controller.go#L477)

and there are code to sync backup to cluster and [code to delete orphaned backups](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_sync_controller.go#L307) as part of this velero upstream sync. So I think this is the same case.

Velero default is [1 minute](https://github.com/vmware-tanzu/velero/blob/6ac38cde8556f0fb338f1b00cd64746790b6c73c/pkg/cmd/server/config/config.go#L24) So our 5 minutes works. Could be as low as 2 min imo.

internal/controller/nonadminbackupsynchronizer_controller.go
```diff
@@ -0,0 +1,168 @@
...
+					},
+				},
+				Spec: nacv1alpha1.NonAdminBackupSpec{
+					BackupSpec: &backup.Spec,
+				},
+			}
+			value, exist := relatedNonAdminBackupStorageLocations[nab.Spec.BackupSpec.StorageLocation]
+			if exist {
+				nab.Spec.BackupSpec.StorageLocation = value
+			}
```
mateusoliveira43, line 148: else, make an empty value

internal/controller/nonadminbackupstoragelocation_controller.go
mateusoliveira43: can not import, its unexported https://github.com/vmware-tanzu/velero/blob/6ac38cde8556f0fb338f1b00cd64746790b6c73c/pkg/cmd/server/config/config.go#L24

internal/controller/nonadminbackup_controller.go
```diff
@@ -117,6 +117,14 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			r.removeNabFinalizerUponVeleroBackupDeletion,
 		}
 
+	case function.CheckLabelAnnotationValueIsValid(nab.Labels, constant.NabSyncLabel):
```
mateusoliveira43, line 120: write test for this scenario

go.mod
```diff
@@ -82,3 +82,5 @@ require (
 )
 
 replace github.com/vmware-tanzu/velero => github.com/openshift/velero v0.10.2-0.20241211163542-fa8f2486175b
+
+replace github.com/openshift/oadp-operator => github.com/mateusoliveira43/oadp-operator v0.0.0-20250205190228-fcc5ea4a6c32
```
mateusoliveira43, line 86: TODO remove prior to merge

shubham-pampattiwar:
Lets address this comment as well https://github.com/migtools/oadp-non-admin/pull/152#issuecomment-2643696352

internal/controller/nonadminbackup_controller.go
```diff
@@ -117,6 +117,14 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		logger.V(1).Info("Executing nab sync path")
+		reconcileSteps = []nonAdminBackupReconcileStepFunction{
```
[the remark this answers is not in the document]
mpryc: Synced backups are intended solely for restore purposes, correct? This means the user can initiate a restore operation but cannot rerun the backup.

In this scenario, skipping validation for synced backups makes sense, as they have already been created and will be used exclusively for restoration only.

internal/controller/nonadminbackup_controller.go
```diff
@@ -117,6 +117,14 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			r.setBackupUUIDInStatus,
+			r.setFinalizerOnNonAdminBackup,
+			r.createVeleroBackupAndSyncWithNonAdminBackup,
```
mpryc, line 125: I suppose the createVeleroBackup will not really create velero backup and will only sync it back to the synced NAB object? Velero Backup is the one triggering sync not vice-versa.

mateusoliveira43: yes, when synced nab reaches that step, it will fetch already created velero backup with its UUID instead of creating new one

mpryc:
Tested:
1)
 A) Created NABSL
 B) Created NAB
 C) Removed NAB
    -> Properly synced NAB

2)
 A) Created NABSL
 B) Created NAB
 C) Modified NABSL to point to improper bucket
 D) Removed NAB
    -> Properly NOT synced NAB

3)
 A) Created NABSL
 B) Created NAB
 C) Modified NABSL to point to improper bucket
 D) Removed NAB
    -> Properly NOT synced NAB
 E) Modified NABSL to point to proper bucket
    -> Properly synced NAB
 
4)
 A) Created NABSL "mybsl"
 B) Created NAB
 C) Deleted non-admin namespace
 D) Created non-admin namespace
 E) ReCreated NABSL "mybsl"
    -> Properly synced NAB
 
5)
 A) Created NABSL "mybsl"
 B) Created NAB
 C) Deleted non-admin namespace
 D) Created non-admin namespace
 E) ReCreated NABSL "mybsltwo" pointing to the same s3 as "mybsl"
    -> Properly synced NAB

 6)
 A) Created NABSL "mybsl"
 B) Created NAB
 C) Deleted non-admin namespace
 D) Created non-admin-two namespace
 E) ReCreated NABSL "mybsl" pointing to the same s3 as "mybsl" in the "non-admin-two" namespace
    -> Properly NOT synced NAB, Velero BSL object has different prefix

PR #139, opened 2025-01-14
Allow Non-Admin user to create Backup Storage Locations - fixes #36

### LINKED ISSUE #36: Ability for non-admin users to create BackupStorageLocations(BSL)

mpryc:
Please note, this PR replaces the https://github.com/migtools/oadp-non-admin/pull/115
I will close the #115 because it was created using operator-sdk and not the kubebuilder as described in the:
 https://github.com/migtools/oadp-non-admin/blob/ca7e290f36a116593ba8255125baf3b8b561dc6c/docs/architecture.md#kubebuilder

## Why the changes were made

Fix #36, also #57

1. Create as usual OADP with NAC.
2. Create user secret in the user's namespace similarly to the S3 secret for DPA:

```shell
$ oc create secret generic nac-credentials -n <USERS_NAMESPACE> --from-file cloud=credentials-velero
```
3. Create Non Admin BSL pointing to the above credential in the user's namespace, example spec part:
```yaml

metadata:
  name: example
  namespace: nactest
spec:
  config:
    region: eu-central-1
  credential:
    key: cloud
    name: nac-credentials
  objectStorage:
    bucket: my-bucket
    prefix: some-prefix
  provider: aws

```

4. Check the NAC logs, the NaBSL object (if it was correctly reconciled and updated with status from Velero BSL)

5...6...7... Do break things, such as improper Secret data, non-existing secret, removal of the secret from the OADP namespace or the user namespace, removal of the BSL from the OADP namespace and see if everything reconciles as expected.

Read [design](https://github.com/mpryc/oadp-non-admin/blob/nonadminstoragelocation_controller/docs/design/non_admin_backupstoragelocation.md)


docs/design/non_admin_backupstoragelocation.md
```diff
@@ -0,0 +1,116 @@
...
+5. Controller creates a Velero BSL resource in the OADP namespace pointing to the Secret from the OADP namespace.
+6. Controller updates the NaBSL Status with the information from the created Velero BSL resource.
+
+### Non-Admin BSL Update Flow
+1. User submits a Non-Admin BSL update request.
+2. Controller validates changes
+3. Controller updates the Secret and/or Velero BSL resource in the OADP namespace based on the Non-Admin BSL UUID.
+4. Controller updates the NaBSL Status with the information from the updated Velero BSL resource.
+
+### Deletion Flow
```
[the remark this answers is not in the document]
mateusoliveira43: need to wait sync

problem is that velero backup will point to non existing velero BSL, if this step is not done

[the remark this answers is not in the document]
mpryc: how to restore from the situation where user will remove all the objects from user namespace? I'd like to walk through steps, could you please write how do you see that?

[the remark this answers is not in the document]
mateusoliveira43: user created NABSL (NAC created BSL in OADP namepasce), then NAB (NAC created backup in OADP namepasce)

user deleted NAB (not from object storage) and NABSL (without garbage collector, may be leftovers on OADP namespace. lets supose backup was deleted from OADP namepasce)

user creates NABSL again. NAC creates BSL in OADP namepasce again (with different name from before, thats why garbage collector can be important). Velero syncs backup in OADP namepasce. NAC syncs NABSL in user namespace. User is ready to restore NAB

api/v1alpha1/nonadminbackupstoragelocation_types.go
```diff
@@ -0,0 +1,104 @@
...
+	NaBSLPhaseNew         NonAdminBackupStorageLocationPhase = "New"
+	NaBSLPhaseAvailable   NonAdminBackupStorageLocationPhase = "Available"
+	NaBSLPhaseUnavailable NonAdminBackupStorageLocationPhase = "Unavailable"
+	NaBSLPhaseCreated     NonAdminBackupStorageLocationPhase = "Created"
+	NaBSLPhaseDeleting    NonAdminBackupStorageLocationPhase = "Deleting"
+)
+
+// NonAdminBSLCondition contains addition conditions to the
+// generic ones defined as NonAdminCondition
+// +kubebuilder:validation:Enum=SecretSynced;BSLSynced
```
mateusoliveira43, line 39: this line should be removed

this type is not referenced in CRD to have the marker

config/samples/oadp_v1alpha1_nonadminbackupstoragelocation.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/name: nonadminbackupstoragelocation
+    app.kubernetes.io/instance: nonadminbackupstoragelocation-sample
+    app.kubernetes.io/part-of: oadp-nac
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: oadp-nac
+  name: nonadminbackupstoragelocation-sample
+spec:
+  # TODO(user): Add fields here
```
mateusoliveira43, line 12: create example

internal/common/function/function.go
```diff
@@ -157,6 +167,82 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
 	return nil
 }
 
+// ValidateBslSpec return nil, if NonAdminRestore is valid; error otherwise
```
mateusoliveira43, line 170: typo

internal/common/function/function.go
```diff
@@ -157,6 +167,82 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
...
+func ValidateBslSpec(ctx context.Context, clientInstance client.Client, nonAdminBsl *nacv1alpha1.NonAdminBackupStorageLocation, enforcedBslSpec *velerov1.BackupStorageLocationSpec) error {
+	if nonAdminBsl.Spec.BackupStorageLocationSpec.Credential == nil {
+		return fmt.Errorf("NonAdminBackupStorageLocation spec.bslSpec.credential is not set")
+	} else if nonAdminBsl.Spec.BackupStorageLocationSpec.Credential.Name == constant.EmptyString || nonAdminBsl.Spec.BackupStorageLocationSpec.Credential.Key == constant.EmptyString {
+		return fmt.Errorf("NonAdminBackupStorageLocation spec.bslSpec.credential.name or spec.bslSpec.credential.key is not set")
+	}
+
+	var enforcedSpec reflect.Value
```
mateusoliveira43, line 178: asked in slack to redo this part the same way as NAB and NAR

internal/common/function/function_test.go
```diff
@@ -688,6 +693,241 @@ func TestValidateRestoreSpecEnforcedFields(t *testing.T) {
 	})
 }
 
+func TestValidateBslSpec(t *testing.T) {
```
mateusoliveira43, line 696: would follow structure of NAB and NAR tests https://github.com/migtools/oadp-non-admin/blob/019bb2a677d4648cce1443966d6686cd46a7cef4/internal/common/function/function_test.go#L94-L377

internal/common/function/function.go
mpryc: correct and my answer is that it can't be the same as here we are dealing with values from ConfigMap which is not the case in other objects, so I am not sure how this can be same?

https://github.com/migtools/oadp-non-admin/pull/139/files/8c8df21394a0a3ea3513d5346ded0c040bfc82c3#diff-de455907ed1c5413fa6248ebea483af559585f8f6277a7c8b370813b9bb254f9R199-R218

internal/common/function/function_test.go
mpryc: I don't understand this request. The difference here is that at the beginning we need to register new corev1 schema and then we create the `tests := [] struct {}` which is pretty much similar to NAB and NAR ?

api/v1alpha1/nonadminbackupstoragelocation_types.go
```diff
@@ -0,0 +1,104 @@
...
+// NonAdminBackupStorageLocationPhase constants similar to velerov1.BackupStorageLocationPhase
+const (
+	NaBSLPhaseNew         NonAdminBackupStorageLocationPhase = "New"
+	NaBSLPhaseAvailable   NonAdminBackupStorageLocationPhase = "Available"
+	NaBSLPhaseUnavailable NonAdminBackupStorageLocationPhase = "Unavailable"
+	NaBSLPhaseCreated     NonAdminBackupStorageLocationPhase = "Created"
+	NaBSLPhaseDeleting    NonAdminBackupStorageLocationPhase = "Deleting"
+)
+
+// NonAdminBSLCondition contains addition conditions to the
```
shubham-pampattiwar, line 37: NIT: additional

api/v1alpha1/nonadminbackupstoragelocation_types.go
```diff
@@ -0,0 +1,104 @@
...
+	// +optional
+	VeleroBackupStorageLocation *VeleroBackupStorageLocation `json:"veleroBackupStorageLocation,omitempty"`
+
+	Phase      NonAdminBackupStorageLocationPhase `json:"phase,omitempty"`
+	Conditions []metav1.Condition                 `json:"conditions,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
```
shubham-pampattiwar, line 83: Lets also add shortname here
`// +kubebuilder:resource:path=nonadminbackupstoragelocations,shortName=nabsl`

cmd/main.go
```diff
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+
```
shubham-pampattiwar, line 33: NIT: remove this

cmd/main.go
```diff
@@ -51,6 +52,11 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+const (
+	unableToCreateControllerString = "unable to create controller"
```
shubham-pampattiwar, line 56: maybe drop the `String` prefix from var names

cmd/main.go
```diff
@@ -153,7 +159,7 @@ func main() {
 		OADPNamespace:      oadpNamespace,
 		EnforcedBackupSpec: enforcedBackupSpec,
 	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "NonAdminBackup")
+		setupLog.Error(err, unableToCreateControllerString, controllerString, "NonAdminBackup")
```
shubham-pampattiwar, line 162: reframe this to `unable to setup <controller-name> controller with manager`

internal/common/constant/constant.go
```diff
@@ -27,20 +27,24 @@ import (
...
+	NabOriginNameAnnotation        = v1alpha1.OadpOperatorLabel + "-nab-origin-name"
+	NabOriginNamespaceAnnotation   = v1alpha1.OadpOperatorLabel + "-nab-origin-namespace"
+	NarOriginNameAnnotation        = v1alpha1.OadpOperatorLabel + "-nar-origin-name"
+	NarOriginNamespaceAnnotation   = v1alpha1.OadpOperatorLabel + "-nar-origin-namespace"
+	NabslOriginNameAnnotation      = v1alpha1.OadpOperatorLabel + "-nabsl-origin-name"
+	NabslOriginNamespaceAnnotation = v1alpha1.OadpOperatorLabel + "-nabsl-origin-namespace"
+
+	NabFinalizerName   = "nonadminbackup.oadp.openshift.io/finalizer"
+	NarFinalizerName   = "nonadminrestore.oadp.openshift.io/finalizer"
+	NabslFinalizerName = "nabsl.oadp.openshift.io/finalizer"
```
shubham-pampattiwar, line 47: lets be uniform here, either use shortname for all the CRs or fullname for all of them in finalizer label.

internal/common/function/function.go
```diff
@@ -157,6 +167,82 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
 	return nil
 }
 
+// ValidateBslSpec return nil, if NonAdminRestore is valid; error otherwise
+func ValidateBslSpec(ctx context.Context, clientInstance client.Client, nonAdminBsl *nacv1alpha1.NonAdminBackupStorageLocation, enforcedBslSpec *velerov1.BackupStorageLocationSpec) error {
+	if nonAdminBsl.Spec.BackupStorageLocationSpec.Credential == nil {
```
shubham-pampattiwar, line 172: Maybe add a comment here with the issue link: https://github.com/migtools/oadp-non-admin/issues/146 as a TODO

internal/common/function/function.go
```diff
@@ -157,6 +167,82 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
...
+				return fmt.Errorf(
+					"NonAdminBackupStorageLocation spec.bslSpec.%v field value is enforced by admin user, cannot override it",
+					tagName,
+				)
+			}
+		}
+	}
+
+	// Check if the secret exists in the same namespace
+	secret := &corev1.Secret{}
```
shubham-pampattiwar, line 233: Should we also check for the `key` specified by the user ?

internal/common/function/function.go
```diff
@@ -427,6 +513,48 @@ func GetVeleroRestoreByLabel(ctx context.Context, clientInstance client.Client,
...
+	secretList := &corev1.SecretList{}
+
+	// Call the generic ListLabeledObjectsInNamespace function
+	if err := ListObjectsByLabel(ctx, clientInstance, namespace, constant.NabslOriginNACUUIDLabel, labelValue, secretList); err != nil {
+		return nil, err
+	}
+
+	switch len(secretList.Items) {
+	case 0:
+		return nil, nil // No matching DeleteBackupRequest found
```
shubham-pampattiwar, line 529: NIT: s/`DeleteBackupRequest`/`secret`

internal/common/function/function.go
```diff
@@ -427,6 +513,48 @@ func GetVeleroRestoreByLabel(ctx context.Context, clientInstance client.Client,
...
+	case 1:
+		return &secretList.Items[0], nil // Found 1 matching DeleteBackupRequest
```
shubham-pampattiwar, line 531: NIT: s/`DeleteBackupRequest`/`secret`

internal/common/function/function_test.go
```diff
@@ -55,9 +55,14 @@ const (
 	testNonAdminRestoreUUID       = "12345678-1234-1234-1234-123456789abc"
 	testNonAdminRestoreNamespace  = "non-admin-restore-namespace"
 	testNonAdminSecondRestoreName = "non-admin-second-restore-name"
-	defaultStr                    = "default"
 	invalidInputEmptyNamespace    = "Invalid input - empty namespace"
-	invalidInputEmptyNsErr        = "invalid input: namespace, labelKey, and labelValue must not be empty"
+	defaultStr                    = "default"
```
shubham-pampattiwar, line 59: maybe drop the `Str` prefix from var names

internal/common/function/function.go
mateusoliveira43: maybe use same validation from OADP, by importing it

internal/common/function/function_test.go
mateusoliveira43: for NAB and NAR the test is split in 2: one for enforcement fields and one for other validations

for enforcement, we also check that every single field was tested https://github.com/migtools/oadp-non-admin/blob/019bb2a677d4648cce1443966d6686cd46a7cef4/internal/common/function/function_test.go#L361

internal/common/function/function.go
mateusoliveira43: the type of the object does not require new implementation

both NAB and NAR also have `map[string]string` fields 

https://github.com/migtools/oadp-non-admin/blob/019bb2a677d4648cce1443966d6686cd46a7cef4/internal/common/function/function_test.go#L285-L289

https://github.com/migtools/oadp-non-admin/blob/019bb2a677d4648cce1443966d6686cd46a7cef4/internal/common/function/function_test.go#L595-L601

internal/common/function/function_test.go
mpryc: @shubham-pampattiwar in go "default" is reserved keyword / because it's used in switch statements, so what else should I change this to?

internal/common/function/function.go
mpryc: Fixed, will commit soon.

internal/common/function/function.go
mpryc: I don't think so. If the user will enter invalid key, then BSL created by the user will propagate the error returning in the NaBSL status error conditions. Duplication of this validation doesn't give us a lot of benefit. Note that also if the credentials are valid, but access to S3 is denied there will be as well error coming from the BSL (Velero) which will be propagated back to us same way through BSL.

internal/common/function/function.go
mpryc: Fixed, will commit soon.

internal/common/constant/constant.go
mpryc: Ok changed it to a very long one as I prefer to be explicit ;)
```go
	NabslFinalizerName = "nonadminbackupstoragelocation.oadp.openshift.io/finalizer"
```
will commit soon.

cmd/main.go
mpryc: Fixed, will commit soon. Removed constant controller and controllerString as well, because those are now static strings.

cmd/main.go
mpryc: Not required anymore as this is currently inline with the error.

cmd/main.go
mpryc: The VSCode is stubborn and it's adding this back, anyway I will need to fix few other imports:

```shell
cmd/main.go:33: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/migtools/oadp-non-admin) (gci)

internal/controller/nonadminbackupstoragelocation_controller.go:24: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/migtools/oadp-non-admin) (gci)

	nacv1alpha1 "github.com/migtools/oadp-non-admin/api/v1alpha1"
	"github.com/migtools/oadp-non-admin/internal/common/function"
	"github.com/migtools/oadp-non-admin/internal/handler"
	"github.com/migtools/oadp-non-admin/internal/predicate"

internal/controller/nonadminbackupstoragelocation_controller.go:33: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/migtools/oadp-non-admin) (gci)

internal/controller/nonadminbackupstoragelocation_controller.go:40: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/migtools/oadp-non-admin) (gci)

internal/controller/nonadminbackupstoragelocation_controller.go:45: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/migtools/oadp-non-admin) (gci)

```

api/v1alpha1/nonadminbackupstoragelocation_types.go
mpryc: Fixed, will commit soon.

internal/common/function/function.go
mpryc: Fixed, will commit soon.

config/samples/oadp_v1alpha1_nonadminbackupstoragelocation.yaml
mpryc: Fixed, will commit soon.

api/v1alpha1/nonadminbackupstoragelocation_types.go
mpryc: Shouldn't matter. We are not exposing this to the user, so it's rather contract with the controller to not set some other values. I will remove that because I don't mind enforcing this one.

cmd/main.go
mateusoliveira43: in `settings.json`
```json
"[go]": {
        "editor.codeActionsOnSave": {
            "source.organizeImports": "never"
        }
    },
```

mpryc: Fixed, will commited. Can we close this as resolved?

api/v1alpha1/nonadminbackupstoragelocation_types.go
mpryc: Can we close this as resolved?

api/v1alpha1/nonadminbackupstoragelocation_types.go
mpryc: Can we close this as resolved?

internal/common/function/function.go
mpryc: Can we close this as resolved?

config/samples/oadp_v1alpha1_nonadminbackupstoragelocation.yaml
mpryc: Can we close this as resolved?

api/v1alpha1/nonadminbackupstoragelocation_types.go
mpryc: Last commit removed that. Can we close this as resolved?

docs/design/non_admin_backupstoragelocation.md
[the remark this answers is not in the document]
mpryc: Can we close this as resolved? I think we agreed on how the Sync controller and NABSL works together?

internal/common/function/function.go
mpryc: I'll look into that.

internal/common/function/function_test.go
mpryc: Thanks for explanation, will look into that.

cmd/main.go
mpryc: Can we close this as resolved?

internal/common/constant/constant.go
mpryc: Can we close this as resolved? or do you prefer shorter names across all? I am fine either way.

internal/common/function/function.go
mpryc: Can we close this as resolved? TODO added and it's pointing 146.

internal/common/function/function.go
mpryc: Can we close this as resolved?

internal/common/function/function.go
mateusoliveira43: you need to pass a `enforcedSpec` to the tests. Function do not expect it to be nil. At max, it will be a empty struct

adding here https://github.com/mpryc/oadp-non-admin/blob/nonadminstoragelocation_controller-v2/internal/common/function/function_test.go#L758

`enforcedSpec: &velerov1.BackupStorageLocationSpec{},` should fix teh panic, for example

mpryc: Are we 100% positive that structure will always be present in the running controller code? I mean if that empty structure is present even if no enforce settings are set by admin ?

mateusoliveira43: yes

https://github.com/migtools/oadp-non-admin/blob/019bb2a677d4648cce1443966d6686cd46a7cef4/cmd/main.go#L201-L213

it is initialized as a pointer to a empty struct

if admin user did set any enforcement, it is updated to a pointer to a non empty struct

mateusoliveira43: again, what you are trying to accomplish goes against design https://github.com/migtools/oadp-non-admin/blob/master/docs/design/admin_control_over_spec.md?plain=1#L27

if you want to change this, you also need to change backup and restore enforcement

mpryc: At the time of writing that design I was not aware of need of the config. I think we must change that. I am not sure if we want to change it for Backup/Restore - if yes I would say separate PR

shubham-pampattiwar: BSL config spec is a bit different that the other non-nested typed spec levels. If we want to revisit backup/restore enforcement specs then lets just make note/create issue for it and move on. It can be a seperate PR. The changes in the current PR cater to NaBSL, lets make it rock solid and keep moving :)

internal/common/function/function_test.go
kaovilai: From checking code, this is only used to convey the default ns name.

I suggests `namespaceDefault` to convey its utility.

kaovilai: `defaultNS` or `defaultNamespace` also works.

internal/common/function/function.go
mpryc: Is the below Issue which will need to be revisited solving the debate within this PR or is change to this PR still required?

https://github.com/migtools/oadp-non-admin/issues/151

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,691 @@
...
+		if updateErr := r.Status().Update(ctx, nabsl); updateErr != nil {
+			logger.Error(updateErr, failedUpdateStatusError)
+			return false, updateErr
+		}
+	}
+
+	return false, nil
+}
+
+// syncVeleroBSLWithNaBSL
```
shubham-pampattiwar, line 624: Lets just say syncStatus*

internal/common/function/function.go
shubham-pampattiwar: We can review and revisit this in #151

internal/common/function/function_test.go
mpryc: changed.

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,691 @@
...
+		}
+	default:
+		// Standard creation/update path
+		logger.V(1).Info("Executing nabsl creation/update path")
+		reconcileSteps = []naBSLReconcileStepFunction{
+			r.initNaBSLCreate,
+			r.validateNaBSLSpec,
+			r.setVeleroBSLUUIDInNaBSLStatus,
+			r.setFinalizerOnNaBSL,
+			r.createSyncNaBSLSecrets,
```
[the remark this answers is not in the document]
mpryc: renamed.

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: renamed.

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+
+	// We use Credential from the secret created in the createSyncNaBSLSecrets function
+	// however we need to set the key to the one specified in the NonAdminBackupStorageLocation spec
+	// because it's the user who decides which key to use from the secret
+	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, veleroBsl, func() error {
+		veleroBsl.Spec.AccessMode = bslSpec.BackupStorageLocationSpec.AccessMode
+		veleroBsl.Spec.BackupSyncPeriod = bslSpec.BackupStorageLocationSpec.BackupSyncPeriod
+		veleroBsl.Spec.Config = bslSpec.BackupStorageLocationSpec.Config
+		veleroBsl.Spec.Credential = &corev1.SecretKeySelector{
```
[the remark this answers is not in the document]
mpryc: Added as issue as I want to test it on the cluster before making such change and I don't want this PR to block others. https://github.com/migtools/oadp-non-admin/issues/153

Also I think I will add the ownerReference between secret and the BSL, because they do share same namespace and it makes sense to remove generated by the cotroller secret when BSL is gone.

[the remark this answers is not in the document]
shubham-pampattiwar: +1

internal/common/function/function.go
mpryc: Added as well to the issue #153 I will double check when I will work on that issue, should be minor, however I prefer to double test it.

shubham-pampattiwar (dismissed):

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,682 @@
...
+			if updateErr := r.Status().Update(ctx, nabsl); updateErr != nil {
+				logger.Error(updateErr, statusBslUpdateError)
+				return false, updateErr
+			}
+		}
+		return false, reconcile.TerminalError(err)
+	}
+
+	// Validation successful, update phase and condition
+	updatedPhase := updateNaBSLPhase(&nabsl.Status.Phase, nacv1alpha1.NaBSLPhaseNew)
```
mateusoliveira43, line 285: this was already done in `initNaBSLCreate`

mpryc: @mateusoliveira43 Fixed

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,682 @@
...
+			).Result()
+	}
+
+	bslSpec := nabsl.Spec.DeepCopy()
+
+	// We use Credential from the secret created in the syncSecrets function
+	// however we need to set the key to the one specified in the NonAdminBackupStorageLocation spec
+	// because it's the user who decides which key to use from the secret
+	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, veleroBsl, func() error {
+		veleroBsl.Spec.AccessMode = bslSpec.BackupStorageLocationSpec.AccessMode
```
mateusoliveira43, line 523: this structure makes it hard to maintain it

why not do the same way as backup and restore?
https://github.com/migtools/oadp-non-admin/blob/25e40d6abd8c49c43a1a89a9f49d6d489d07707e/internal/controller/nonadminbackup_controller.go#L620-L621

https://github.com/migtools/oadp-non-admin/blob/25e40d6abd8c49c43a1a89a9f49d6d489d07707e/internal/controller/nonadminrestore_controller.go#L344-L346

only field that needs to change in BSL spec compared to NABSL is credential, not?

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,679 @@
...
+		return false
+	}
+
+	// Update and return true if they differ
+	status.VeleroBackupStorageLocation.Status = veleroBackupStorageLocation.Status.DeepCopy()
+	return true
+}
+
+// updateNaBSLPhase updates the phase of the NonAdminBackupStorageLocation
+func updateNaBSLPhase(phase *nacv1alpha1.NonAdminBackupStorageLocationPhase, newPhase nacv1alpha1.NonAdminBackupStorageLocationPhase) bool {
```
mateusoliveira43, line 667: what is the difference between this and `updateNonAdminBslPhase`?

api/v1alpha1/nonadminbackupstoragelocation_types.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// NonAdminBackupStorageLocationPhase is a simple one high-level summary of the lifecycle of an NonAdminBackupStorageLocation.
+// +kubebuilder:validation:Enum=New;Available;Unavailable;Created;Deleting
+type NonAdminBackupStorageLocationPhase string
+
+// NonAdminBackupStorageLocationPhase constants similar to velerov1.BackupStorageLocationPhase
+const (
+	NaBSLPhaseNew         NonAdminBackupStorageLocationPhase = "New"
+	NaBSLPhaseAvailable   NonAdminBackupStorageLocationPhase = "Available"
```
mateusoliveira43, line 31: is this phase used?

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -0,0 +1,679 @@
...
+func (r *NonAdminBackupStorageLocationReconciler) validateNaBSLSpec(ctx context.Context, logger logr.Logger, nabsl *nacv1alpha1.NonAdminBackupStorageLocation) (bool, error) {
+	// Skip validation if not in New phase
+	if nabsl.Status.Phase != nacv1alpha1.NaBSLPhaseNew {
+		logger.V(1).Info("Skipping validation, not in New phase", constant.CurrentPhaseString, nabsl.Status.Phase)
+		return false, nil
+	}
+
+	err := function.ValidateBslSpec(ctx, r.Client, nabsl)
+	if err != nil {
+		updatedPhase := updateNonAdminBslPhase(&nabsl.Status.Phase, nacv1alpha1.NaBSLPhaseUnavailable)
```
mateusoliveira43, line 266: can this confuse users?

Velero BSL in unavailable state means it has not been able to connect to bucket, for example

in our case, BSL was not even created if it fails this step

internal/controller/nonadminbackupstoragelocation_controller_test.go
```diff
@@ -0,0 +1,466 @@
...
+	})
+
+	ginkgo.DescribeTable("Validation is false",
+		func(scenario nonAdminBackupStorageLocationClusterValidationScenario) {
+			nonAdminBsl := buildTestNonAdminBackupStorageLocation(nonAdminBslNamespace, nonAdminBslName, scenario.spec)
+			err := k8sClient.Create(ctx, nonAdminBsl)
+			gomega.Expect(err).To(gomega.HaveOccurred())
+			gomega.Expect(err.Error()).To(gomega.ContainSubstring("Required value"))
+		},
+		ginkgo.Entry("Should NOT create NonAdminBackupStorageLocation without spec.restoreSpec", nonAdminBackupStorageLocationClusterValidationScenario{
```
mateusoliveira43, line 167: typo

internal/controller/nonadminbackupstoragelocation_controller_test.go
```diff
@@ -0,0 +1,466 @@
...
+			spec: nacv1alpha1.NonAdminBackupStorageLocationSpec{},
+		}),
+	)
+})
+
+var _ = ginkgo.Describe("Test full reconcile loop of NonAdminRestore Controller", func() {
```
mateusoliveira43, line 173: typo

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: I did not want to introduce new phase and keep it 1-1 same as BSL.

This one is unavailable, but with the condition telling why it's not available (next line)

shubham-pampattiwar: Its fine NAC objects can have their own status phases and interpretation, right ?

mpryc: I will change this to be aligned with phases from Backup and Restore, so they are same and user won't be confused:
https://github.com/migtools/oadp-non-admin/blob/25e40d6abd8c49c43a1a89a9f49d6d489d07707e/api/v1alpha1/nonadmin_types.go#L23-L32

api/v1alpha1/nonadminbackupstoragelocation_types.go
mpryc: Please see my comment below, I will adjust those phases to the:
https://github.com/migtools/oadp-non-admin/blob/25e40d6abd8c49c43a1a89a9f49d6d489d07707e/api/v1alpha1/nonadmin_types.go#L23-L32

The initial idea was to match the BSL but the BSL satus is fully reflected in the Status anyway so it makes sense to align with other NAC objects

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: I will use one from nonadminbackup_controller.go. Ideally that function syould be part of function.go, but that's not something I will do in this PR as this will modify other controllers which I don't want to perform in this PR

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: Changed. Note this is slightly different and we need to use CreateOrUpdate function here. In the Restore or Backup situation user can not modify created object as it will not be reflected on the final object state (sync controller will get the backup from the s3 for example and the user's changes are overriden).

This is NOT the case in the BSL scenario, where the user can modify settings such as credentials and they will get updated and reflected in the Velero BSL case, so we should behave similarly.

internal/controller/nonadminbackupstoragelocation_controller_test.go
mpryc: Fixed.

internal/controller/nonadminbackupstoragelocation_controller_test.go
mpryc: Fixed.

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: > Its fine NAC objects can have their own status phases and interpretation, right ?

I have changed current implementation to align with `Backup` and `Restore` phases so now NAC have one set of phases for all objects.

mpryc:
@mateusoliveira43 I think most of the comments are addressed, please let me know if any further changes are required. Thanks.

internal/controller/nonadminbackupstoragelocation_controller_test.go
```diff
@@ -0,0 +1,466 @@
...
+				},
+				nonAdminBsl,
+			)).To(gomega.Succeed())
+
+			ginkgo.By("Validating NonAdminBackupStorageLocation Status")
+
+			gomega.Expect(checkTestNonAdminBackupStorageLocationStatus(nonAdminBsl, scenario.expectedStatus)).To(gomega.Succeed())
+
+			veleroBsl := &velerov1.BackupStorageLocation{}
+			if scenario.expectedStatus.VeleroBackupStorageLocation != nil && len(nonAdminBsl.Status.VeleroBackupStorageLocation.NACUUID) > 0 {
```
mateusoliveira43, line 272: this is not being run, need to add
```go
                                VeleroBackupStorageLocation: &nacv1alpha1.VeleroBackupStorageLocation{
					Status: nil,
				},
```
here https://github.com/mpryc/oadp-non-admin/blob/cc9b49bacce0286986c1bc5ef3ef5986ce06163b/internal/controller/nonadminbackupstoragelocation_controller_test.go#L438

PR #148, opened 2025-01-23
Modify the Non Admin Backup API deletion to align with sync controller

mpryc:
Non Admin Backup delete event should remove corresponding Velero Backup

This allows Velero Sync controller to recreate Velero Backup object and cascate such recreation back to the Non Admin Backup. The cascade action is a responsibility of an Non Admin Sync controller.

## Why the changes were made

Those changes are made to allow sync controller that is implemented as part of the #38 implemented by #138 to recreate Non Admin Backup in unified way, meaning it will be recreated on the new Velero Backup recreation by the Velero Sync controller.

In other words with this behavior when user runs:
```shell
$ oc delete <non-admin-backup> -n mynamespace
```

This will result in removal of associated Velero Backup (if was existing) and then removal of NAB finalizer.
Such action will cause s3 storage to still have the information about the Velero Backup, because the removal was not done via spec field and no DeleteBackupRequest was created.
With the s3 storage containing information about Velero Backup, the Velero Sync Controller will recreate the Velero Backup object in the `openshift-adp` namespace. That event should trigger Non Admin Backup Sync Controller implemented as part of the #138 to recreate Non Admin Backup in the user's namespace.

e2e tests were written, run simulation tests

### DIFF
--- a/internal/controller/nonadminbackup_controller.go
+++ b/internal/controller/nonadminbackup_controller.go
@@ -94,20 +94,6 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	// First switch statement takes precedence over the next one
 	switch {
-	case nab.Spec.ForceDeleteBackup:
-		// Force delete path - immediately removes both VeleroBackup and DeleteBackupRequest
-		// Remove dependent VeleroBackup object
-		// Remove finalizer from the NonAdminBackup object
-		// If there was existing BSL pointing to the Backup object
-		// the Backup will be restored causing the NAB to be recreated
-		logger.V(1).Info("Executing force delete path")
-		reconcileSteps = []nonAdminBackupReconcileStepFunction{
-			r.setStatusAndConditionForDeletionAndCallDelete,
-			r.deleteVeleroBackupObjects,
-			r.deleteDeleteBackupRequestObjects,
-			r.removeNabFinalizerUponVeleroBackupDeletion,
-		}
-
 	case nab.Spec.DeleteBackup:
 		// Standard delete path - creates DeleteBackupRequest and waits for VeleroBackup deletion
 		logger.V(1).Info("Executing standard delete path")
@@ -127,6 +113,7 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		reconcileSteps = []nonAdminBackupReconcileStepFunction{
 			r.setStatusForDirectKubernetesAPIDeletion,
 			r.deleteVeleroBackupObjects,
+			r.deleteDeleteBackupRequestObjects,
 			r.removeNabFinalizerUponVeleroBackupDeletion,
 		}
 
@@ -263,7 +250,6 @@ func (r *NonAdminBackupReconciler) createVeleroDeleteBackupRequest(ctx context.C
 	}
 
 	// Initiate deletion of the VeleroBackup object only when the finalizer exists.
-	// For the ForceDelete we do not create DeleteBackupRequest
 	veleroBackupNACUUID := nab.Status.VeleroBackup.NACUUID
 	veleroBackup, err := function.GetVeleroBackupByLabel(ctx, r.Client, r.OADPNamespace, veleroBackupNACUUID)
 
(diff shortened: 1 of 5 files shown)


internal/controller/nonadminbackup_controller.go
```diff
@@ -96,10 +96,15 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	switch {
 	case nab.Spec.ForceDeleteBackup:
```
mateusoliveira43, line 97: should we remove this option? It basically does same thing as normal delete process now

Removing it can avoid customers confusion and simplify code logic

mpryc: I will ask @shubham-pampattiwar for 3rd opinion as I thought we agree to make this valid option to ensure that DeleteBackupRequest is also removed. Standard delete nor oc delete won't remove this object if something goes wrong.

kaovilai (commented):
Just wanna add that another mechanism you should consider is OwnerReferences.

https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

Perhaps you can patch NABackup to be owned by VBackup. When VBackup is deleted by deleteBackupRequests, k8s will delete NABackup automatically. A finalizer can be added to NABackup to perform some cleanup tasks before letting deletion go through.

This is the same mechanism that delete any pods associated with deployment when deployment is deleted.

mpryc:
> Just wanna add that another mechanism you should consider is OwnerReferences.
> 
> https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
> 
> Perhaps you can patch NABackup to be owned by VBackup. When VBackup is deleted by deleteBackupRequests, k8s will delete NABackup automatically. A finalizer can be added to NABackup to perform some cleanup tasks before letting deletion go through.

Unfortunately we can not. I was considering this, however it won't work between two different namespaces. Remember that the NABackup and VBackup are in two distinct namespaces.

See in the link you provided:
```
Note:
Cross-namespace owner references are disallowed by design.
```

internal/controller/nonadminbackup_controller.go
mpryc: @mateusoliveira43 if we want delete this option then I will move part to delete `DeleteBackupRequest` under standard `oc delete <nabsl>` workflow as I think it's needed to delete that artefact if for some reason it fails.

This will give user only two options:
- `Spec.deleteBackup` which will create `DeleteBackupRequest` and remove finalizer once everything is gone
- `oc delete <nab>` it will simply remove all the objects causing them to be recreated by Velero Sync Controller and our new controller, however it will also remove all the objects from a "bad" state including all referenced `DeleteBackupRequest` objects, which can be caused by various external conditions such as interrupted call to remove `s3 data`. The second part may or may not recreate `VeleroBackup` and `NABackup` depending on what actually happened before, that's why I still think it would be more feasible to add as part of Force deletion, but I am OK to moving it to delete.

mateusoliveira43: I would not care about leftover `DeleteBackupRequests` objects. As I said in another PR, Velero itself garbage collects them

internal/controller/nonadminbackup_controller_test.go
```diff
@@ -365,6 +366,17 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
 				)
 				gomega.Expect(errors.IsNotFound(err)).To(gomega.BeTrue())
 			}
+
+			veleroBackup := &velerov1.Backup{}
```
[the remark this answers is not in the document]
mateusoliveira43: this would be a E2E test. To test DeleteBackupRequest we need Velero running. In these tests we only mock Velero interactions

internal/controller/nonadminbackup_controller.go
mpryc: @mateusoliveira43 you've convinced me. I've removed the force delete from the Spec field. Note that I've moved as well the deletion of the `DownloadBackupRequest` to the standard API delete. This is to ensure the object is removed cleanly and if it fails, then the Velero itself will clean those objects for us using it's garbage collector.

internal/controller/nonadminbackup_controller_test.go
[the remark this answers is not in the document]
mpryc: @mateusoliveira43 Fixed in the last commit. Please let me know if this PR requires further work if not I will prepare one for the oadp-operator spec, because we just removed ForceDelete which needs to be also there.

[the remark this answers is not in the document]
mateusoliveira43: seems ok to me

mpryc:
All the jobs are passing, is there anything else I should change in this PR to get it merged?

mateusoliveira43 (approved):

shubham-pampattiwar (approved):
/lgtm

PR #155, opened 2025-01-29
Enforcement implementation rework

mpryc:
## Why the changes were made

Draft to implement https://github.com/migtools/oadp-non-admin/pull/156


mpryc:
Different approach taken to make it less generic and enforce entire structures instead of underlying structure values. Resolved in: #236

Issue #151, opened 2025-01-24
Support for second-level enforcement in the NonAdminBackup and NonAdminRestore specifications


mpryc:
Problem statement:
The current design and implementation for admin user enforcement of the spec values in `NonAdminBackup` and `NonAdminRestore` do not support second-level configuration options:

https://github.com/migtools/oadp-non-admin/blob/019bb2a677d4648cce1443966d6686cd46a7cef4/docs/design/admin_control_over_spec.md?plain=1#L27

It is necessary to review the need for second-level enforcement for these objects and adjust the design and implementation accordingly if required.

As an example, the `NonAdminBackupStorageLocation` object requires second-level enforcement to allow the admin user to restrict users to a specific S3 region while still permitting them to use their own buckets within that region. This enforcement must be implemented through the second-level configuration in the spec. This is explained in the: https://github.com/migtools/oadp-non-admin/pull/139#discussion_r1929085746

mpryc:
@shubham-pampattiwar @mateusoliveira43 @weshayutin 

We should revisit restrictions and enforcement for Non Admin Objects.

While looking at our discussion around second-level enforcement of the NABSL I looked into that area and I think we have to double check if any of the values which user may set won't allow user to access data which the user should not have permissions to access. An example would be the [Backup Hook](https://github.com/vmware-tanzu/velero/blob/294bbbc69ed8488e76c33648df35e075a78e346c/pkg/apis/velero/v1/backup_types.go#L124) which uses [Resouces](https://github.com/vmware-tanzu/velero/blob/294bbbc69ed8488e76c33648df35e075a78e346c/pkg/apis/velero/v1/backup_types.go#L197) that are in fact allowing to use [IncludedNamespaes](https://github.com/vmware-tanzu/velero/blob/294bbbc69ed8488e76c33648df35e075a78e346c/pkg/apis/velero/v1/backup_types.go#L210) or even going further they do allow to use [PreHooks](https://github.com/vmware-tanzu/velero/blob/294bbbc69ed8488e76c33648df35e075a78e346c/pkg/apis/velero/v1/backup_types.go#L236) that actually allows to define any [executable](https://github.com/vmware-tanzu/velero/blob/294bbbc69ed8488e76c33648df35e075a78e346c/pkg/apis/velero/v1/backup_types.go#L259) in a process which is owned by Velero pod not a non-admin user. This could potentially result in users performing actions with elevated privileges.


Here is my suggestion which we should take action on:

[...]

PR #156, opened 2025-01-29
Updated enforcement design to align with the issue 151

mpryc:
Implementing enforcement for the NonAdminBackupStorageLocation requires revisit of the enforcement behavior. This design should align with it's requirements and allow other parts of the enforcement functionality.

The enforcement levels are described in the:
https://github.com/migtools/oadp-non-admin/issues/151

### DIFF
--- a/docs/design/admin_control_over_spec.md
+++ b/docs/design/admin_control_over_spec.md
@@ -2,39 +2,53 @@
 
 ## Abstract
 
-Non Admin Controller (NAC) restricts the usage of OADP operator with NonAdminBackups and NonAdminRestores.
-Admin users may want to further restrict this by restricting NonAdminBackup/NonAdminRestore spec fields values.
-
+Non Admin Controller (NAC) restricts the usage of OADP operator with NonAdminBackups, NonAdminRestores and NonAdminBackupStorageLocations.
+Admin users may want to further restrict this by restricting relevant spec fields values.
 ## Background
 
-Non Admin Controller (NAC) adds the ability to admin users restrict the use of OADP operator for non admin users, by only allowing them to create backup/restores from their namespaces with NonAdminBackups/NonAdminRestores.
-Admin users may want to further restrict non admin users operations, like forcing a specific time to live (TTL) for NonAdminBackups associated Velero Backups.
-This design enables admin users to set custom default values for NonAdminBackup/NonAdminRestore spec fields, which can not be overridden by non-admin users.
+The Non Admin Controller (NAC) restricts how non-admin users interact with the OADP operator. It allows them to create backups, restores, and backup storage locations only within their own namespaces using `NonAdminBackup`, `NonAdminRestore`, and `NonAdminBackupStorageLocation`, while enforcing additional restrictions and policies.
+
+The NAC Controller enforces two types of controls: **restrictions** and **enforcements**.
+
+- **Controller Restrictions** define policies that neither users nor cluster administrators can override.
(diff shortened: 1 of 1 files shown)


docs/design/admin_control_over_spec.md
```diff
@@ -2,49 +2,63 @@
...
+
+- **Controller Restrictions** define policies that neither users nor cluster administrators can override.
+- **Controller Enforcements** define policies that administrators can override if needed.
+
+This design allows system administrators to manage **Controller Enforcements**, enabling them to set custom default values for the `NonAdminBackup`, `NonAdminRestore`, and `NonAdminBackupStorageLocation` spec fields, which non-admin users cannot override.
+
+For example, administrators may want to enforce specific rules on non-admin user operations, such as:
+
+- Enforcing a specific time to live (TTL) for Velero Backups associated with `NonAdminBackup`.
+- Limiting the region that users can select within Velero `BackupStorageLocation`.
```
kaovilai, line 21: What benefit does limiting region bring to admins exactly?

mpryc: - AWS pricing varies by region. Admins can enforce the use of cost-effective regions and avoid expensive cross-region data transfer fees.
- Performance optimization to ensure s3 region is closer to the cluster
- Simplified management by ensuring everything is in one regions so admins can review users usage based on one region, they can better monitor access logs, apply regional security policies as everything is bound to one region
- I may also think that GDPR or HIPAA may take place where some data must be stored in a specific locations, e.g. in Poland you can not store any medical data outside of physical location data centers which must be based in the EEA (EU countries + Norway, Iceland, Liechtenstein):

```text
GDPR (General Data Protection Regulation)
[...]
Article 44-49: Transfers of personal data outside the European Economic Area (EEA) (EU + Norway, Iceland, Liechtenstein) are restricted unless specific conditions are met.
```

So above in AWS, Azure, Google Cloud - you must ensure data is stored in an EEA-based region (e.g., AWS Frankfurt, AWS Stockholm).

kaovilai: So we are assuming non admins who brought their own creds are actually sharing admin budgets constraints..

I guess up to admin to decide.

kaovilai: > Performance optimization to ensure s3 region is closer to the cluster

enforcement error/warn should say what is allowed then

mpryc: Is the line 32 for this?

https://github.com/migtools/oadp-non-admin/pull/156/files#diff-788dfa1caf04072e867eaa6c490201f32d2e8bdf4cda035ba397d4d974737850R32

kaovilai: Sure. Just the background was unclear to me. Perhaps adding above bullet points to background would be great.

mpryc:
@shubham-pampattiwar Is this design something we want to work on/modify/close without implementation later?

shubham-pampattiwar:
> @shubham-pampattiwar Is this design something we want to work on/modify/close without implementation later?

@mpryc We want this for Self-service FC

weshayutin:
6 days from feature complete and we still have open design?

mpryc:
Closing.

The enforcement is working for all the Non Admin object the same way, so there is no need to update design to adapt new logic.

This was agreed and we do not want to have different enforcement mechanism for sub level values like BSL Config's at this moment.

PR #159, opened 2025-01-30
Enable Usage of NABSL in NAB object

### LINKED ISSUE #116: Enable Usage of NABSL in NAB object

mpryc:
Allows to use NaBSL inside Non Admin Backup object

The Backup object must reference NABSL object
and not the BSL from Velero.

## Why the changes were made

To fix #116 

1. make simulation tests
2. create NaBSL and reference it in the NaBackup

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -89,7 +89,7 @@ func containsOnlyNamespace(namespaces []string, namespace string) bool {
 }
 
 // ValidateBackupSpec return nil, if NonAdminBackup is valid; error otherwise
-func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup, enforcedBackupSpec *velerov1.BackupSpec) error {
+func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpNamespace string, nonAdminBackup *nacv1alpha1.NonAdminBackup, enforcedBackupSpec *velerov1.BackupSpec) error {
 	if nonAdminBackup.Spec.BackupSpec.IncludedNamespaces != nil {
 		if !containsOnlyNamespace(nonAdminBackup.Spec.BackupSpec.IncludedNamespaces, nonAdminBackup.Namespace) {
 			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces can not contain namespaces other than: %s", nonAdminBackup.Namespace)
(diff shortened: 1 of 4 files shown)


internal/common/function/function.go
```diff
@@ -101,6 +101,37 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup, enforcedBack
 		}
 	}
 
+	// We do this validation again just before we create the VeleroBackup object
```
mateusoliveira43, line 104: I would not run it twice

this is run prior to creating NAB, something would need to change very quickly to pass first validation and fail second

internal/controller/nonadminbackup_controller.go
```diff
@@ -640,6 +643,42 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
 			Spec: *backupSpec,
 		}
 
+		if backupSpec.StorageLocation != constant.EmptyString {
```
mateusoliveira43, line 646: move this up to line 635 to be alongside `backupSpec.IncludedNamespaces = []string{nab.Namespace}`

mpryc: We need line 636 before this if, because on line 679 we use object from 636:
```
			veleroBackup.Spec.StorageLocation = veleroBackupStorageLocation.Name
```

internal/common/function/function.go
shubham-pampattiwar: yeah I agree with @mateusoliveira43 
If there is anything wrong with the BSL the error would be cascaded back to the NAB object status. So we are covered there.

internal/controller/nonadminbackup_controller.go
mpryc: Actually I can assign to pointer directly, so will move that to the 635

internal/common/function/function.go
mpryc: Ok, thanks will  update

internal/controller/nonadminbackup_controller.go
mpryc: Should be fixed now.

internal/common/function/function.go
mpryc: Simplified the logic and now the validations are ran once. Of course we are getting object to find out which BSL are we dealing with to include that in the resulting Velero Backup Object, but that happens without further validations and we are now just returning errors if k8s API gives an error.

internal/controller/nonadminbackup_controller.go
```diff
@@ -630,6 +628,27 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
 			}
 		}
 
+		// Included Namespaces are set by the controller and can not be overridden by the user
+		// nor admin user
+		backupSpec.IncludedNamespaces = []string{nab.Namespace}
+		if backupSpec.StorageLocation != constant.EmptyString {
```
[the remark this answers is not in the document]
mpryc: You are right we already checked for existence and we can safely use that from Status, will change.

internal/common/function/function.go
```diff
@@ -101,6 +101,33 @@ func ValidateBackupSpec(nonAdminBackup *nacv1alpha1.NonAdminBackup, enforcedBack
...
+		err := clientInstance.Get(ctx, types.NamespacedName{
+			Name:      nonAdminBackup.Spec.BackupSpec.StorageLocation,
+			Namespace: nonAdminBackup.Namespace,
+		}, nonAdminBsl)
+		if apierrors.IsNotFound(err) {
+			return fmt.Errorf("NonAdminBackupStorageLocation not found in the namespace: %v", err)
+		} else if err != nil {
+			return fmt.Errorf("NonAdminBackup spec.backupSpec.storageLocation is invalid: %v", err)
+		}
+		if nonAdminBsl.Status.VeleroBackupStorageLocation == nil || nonAdminBsl.Status.VeleroBackupStorageLocation.NACUUID == constant.EmptyString {
```
mateusoliveira43, line 115: here would not be simpler to just check `nonAdminBsl.Status.VeleroBackupStorageLocation.Status.Phase`?

mpryc: we do this later to compare if the phase is available, but I wonder if it may happen that UUID is empty and that's the check for?

mateusoliveira43 (approved):
did not test in cluster, code seems ok

shubham-pampattiwar (approved):
changes look good ! 
Worked fine for me ! 
Screenshot 2025-01-30 at 11 46 03 AM

Issue #146, opened 2025-01-22
Introduce NaBSL validations


shubham-pampattiwar:
- ~~Unique prefix is needed~~ (https://github.com/migtools/oadp-non-admin/pull/172)
- ~~NaBSL cannot be a default BSL~~ (fixed via https://github.com/migtools/oadp-non-admin/pull/161)
- ~~NaBSL creation is necessary or not (needs discussion)~~ (Admins can create a default BSL to be shared amongst the non-admin users)

PR #161, opened 2025-01-30
Restrict NonAdminBSL to set default flag

mpryc:
Partial implementation for the #146

## Why the changes were made

To restrict default flag from being used for the NaBSL objects.

Run tests
```shell
$ make simulation-tests
```

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -171,6 +171,9 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
 func ValidateBslSpec(ctx context.Context, clientInstance client.Client, nonAdminBsl *nacv1alpha1.NonAdminBackupStorageLocation) error {
 	// TODO Introduce validation for NaBSL as described in the
 	// https://github.com/migtools/oadp-non-admin/issues/146
+	if nonAdminBsl.Spec.BackupStorageLocationSpec.Default {
+		return fmt.Errorf("NonAdminBackupStorageLocation can not set spec.bslSpec.default to true")
+	}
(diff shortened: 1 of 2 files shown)


internal/common/function/function.go
```diff
@@ -171,6 +171,9 @@ func ValidateRestoreSpec(ctx context.Context, clientInstance client.Client, nonA
 func ValidateBslSpec(ctx context.Context, clientInstance client.Client, nonAdminBsl *nacv1alpha1.NonAdminBackupStorageLocation) error {
 	// TODO Introduce validation for NaBSL as described in the
 	// https://github.com/migtools/oadp-non-admin/issues/146
+	if nonAdminBsl.Spec.BackupStorageLocationSpec.Default {
+		return fmt.Errorf("NonAdminBackupStorageLocation can not set spec.bslSpec.default to true")
```
mateusoliveira43, line 175: its not `bslSpec` https://github.com/migtools/oadp-non-admin/blob/b5126fddf738c5f3155b902b9a4981a47aee15dc/api/v1alpha1/nonadminbackupstoragelocation_types.go#L36

and would rewrite message to something like this
```
NonAdminBackupStorageLocation spec.backupStorageLocationSpec.default can not be set to true
```

shubham-pampattiwar: +1

mpryc: changed to:
`NonAdminBackupStorageLocation cannot be used as a default BSL`

shubham-pampattiwar (approved):

mateusoliveira43 (approved):

PR #162, opened 2025-01-31
Remove associated Velero Backups after BSL removal

mpryc:
- adds label to the Velero Backup object from the origin NaBSL
 - cleans up the Velero Backup objects with the associated label

## Why the changes were made

Removal of the Backup objects after NaBSL removal should happen from the non admin BSL reconcile.

If there are any left-overs GC should take action, so the GC controller should be independent of this logic.

Responsibility to clean up Non Admin Backups when the Velero Backups objects are removed is GC.

1. Create NaBSL
2. Associate the Backup with the NaBSL
3. Remove the NaBSL
4. Ensure the Velero Backup is deleted

### DIFF
--- a/docs/design/non_admin_backupstoragelocation.md
+++ b/docs/design/non_admin_backupstoragelocation.md
@@ -28,15 +28,19 @@ flowchart TD
 
     %% Delete Flow
     OPERATION -->|**Delete**| SET_PHASE_DELETING[Set Phase: Deleting]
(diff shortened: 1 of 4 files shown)


mpryc:
After looking how this works, I think this PR is wrong.

The NaBSL removal should trigger Non Admin Backup removals and not the Velero Backup removals.

Then Non Admin Backup controller responsibility is to remove the Velero Backup from the `openshift-adp` namespace.

mpryc:
Closing as the implementation in the https://github.com/migtools/oadp-non-admin/pull/163 is used.

PR #163, opened 2025-01-31
Remove associated Non Admin Backups during Non Admin BSL removal

### LINKED ISSUE #147: Cleanup action when NaBSL is deleted by the user
The NaBSL controller should:
- Remove the corresponding BSL object 
- Remove the NAB objects referencing the NaBSL 
- The Velero Objects will be deleted by NAB controller
- Any residual/orphaned Velero Objects will be cleaned up by NaGC controller

mpryc:
Fixes #147 

Calls delete on the Non Admin Backup objects referencing Non Admin BSL being removed. The Non Admin Backup controller will propagate the deletion further to remove Velero Backup objects associated with the Non Admin Backups.

## Why the changes were made

Removal of the Non Admin Backup objects after NaBSL removal should happen from the non admin BSL reconcile.

If there are any left-overs GC should take action, so the GC controller should be independent of this logic.

The following delete call will trigger NAB reconcile to propagate this further and remove the Velero Backups
associated with the Non Admin Backup objects being removed.

1. Create NaBSL
2. Associate the Non Admin Backup with the NaBSL
3. Remove the NaBSL
4. Ensure the Non Admin Backup is deleted
5. Ensure the Velero Backup is deleted (this should happen via NAB controller)

### DIFF
--- a/docs/design/non_admin_backupstoragelocation.md
+++ b/docs/design/non_admin_backupstoragelocation.md
@@ -28,15 +28,19 @@ flowchart TD
 
     %% Delete Flow
     OPERATION -->|**Delete**| SET_PHASE_DELETING[Set Phase: Deleting]
-    SET_PHASE_DELETING --> CHECK_SECRET_EXISTS{Check if Secret Exists}
+    SET_PHASE_DELETING --> CHECK_SECRET_EXISTS{Check if Secret for the NaBSL Exists}
     CHECK_SECRET_EXISTS -->|Yes| DELETE_SECRET[Delete Secret in OADP Namespace]
-    CHECK_SECRET_EXISTS -->|No| CHECK_BSL_EXISTS{Check if Velero BSL Exists}
+    CHECK_SECRET_EXISTS -->|No| CHECK_BSL_EXISTS{Check if Velero BSL for the NaBSL Exists}
 
     DELETE_SECRET --> CHECK_BSL_EXISTS
     CHECK_BSL_EXISTS -->|Yes| DELETE_BSL[Delete Velero BSL Resource in OADP Namespace]
-    CHECK_BSL_EXISTS -->|No| REMOVE_FINALIZER[Remove Finalizer from NaBSL Resource]
+    CHECK_BSL_EXISTS -->|No| CHECK_BACKUPS_EXISTS{Check if NonAdminBackups for the NaBSL Exists}
 
-    DELETE_BSL --> REMOVE_FINALIZER
+    DELETE_BSL --> CHECK_BACKUPS_EXISTS
+    CHECK_BACKUPS_EXISTS -->|Yes| DELETE_BACKUPS[Delete NonAdminBackups in the user's namespace]
+    CHECK_BACKUPS_EXISTS -->|No| REMOVE_FINALIZER[Remove Finalizer from NaBSL Resource]
+
+    DELETE_BACKUPS --> REMOVE_FINALIZER
 
     %% Endpoints
     INVALID_CONFIG --> END[End Reconciliation]
@@ -61,6 +65,8 @@ flowchart TD
     DELETE_SECRET
     CHECK_BSL_EXISTS
     DELETE_BSL
+    CHECK_BACKUPS_EXISTS
+    DELETE_BACKUPS
     REMOVE_FINALIZER
     end
 
@@ -72,8 +78,8 @@ flowchart TD
 
     %% Apply styles
     class START,END endpoint
-    class OPERATION,VALIDATE_CONFIG,CHECK_SECRET_EXISTS,CHECK_BSL_EXISTS decision
-    class GENERATE_UUID,CREATE_OR_UPDATE_SECRET,CREATE_OR_UPDATE_BSL,DELETE_SECRET,DELETE_BSL,REMOVE_FINALIZER process
+    class OPERATION,VALIDATE_CONFIG,CHECK_SECRET_EXISTS,CHECK_BSL_EXISTS,CHECK_BACKUPS_EXISTS decision
+    class GENERATE_UUID,CREATE_OR_UPDATE_SECRET,CREATE_OR_UPDATE_BSL,DELETE_SECRET,DELETE_BSL,DELETE_BACKUPS,REMOVE_FINALIZER process
     class INVALID_CONFIG,UPDATE_STATUS,VALID_CONFIG,SET_PHASE_DELETING,SET_PHASE_NEW,SET_PHASE_CREATED phase
 ```
 
@@ -118,4 +124,5 @@ flowchart TD
 1. User deletes the Non-Admin BSL resource.
 2. Controller deletes the Secret from the OADP namespace based on the Non-Admin BSL UUID.
 3. Controller deletes the Velero BSL resource from the OADP namespace based on the Non-Admin BSL UUID.
-4. Controller removes the finalizer from the Non-Admin BSL resource.
+4. Controller deletes the NonAdminBackups for the NaBSL from the user's namespace based on the Non-Admin BSL UUID.
+5. Controller removes the finalizer from the Non-Admin BSL resource.
--- a/docs/design/non_admin_backupstoragelocation.md
+++ b/docs/design/non_admin_backupstoragelocation.md
@@ -37,11 +37,12 @@ flowchart TD
     CHECK_BSL_EXISTS -->|No| CHECK_BACKUPS_EXISTS{Check if NonAdminBackups for the NaBSL Exists}
 
     DELETE_BSL --> CHECK_BACKUPS_EXISTS
-    CHECK_BACKUPS_EXISTS -->|Yes| DELETE_BACKUPS[Delete NonAdminBackups in the user's namespace]
+    CHECK_BACKUPS_EXISTS -->|Yes| SET_DELETE_BACKUPS[Call delete on the NonAdminBackup object]
     CHECK_BACKUPS_EXISTS -->|No| REMOVE_FINALIZER[Remove Finalizer from NaBSL Resource]
 
-    DELETE_BACKUPS --> REMOVE_FINALIZER
-
+    SET_DELETE_BACKUPS --> |"delete call failed"| REQUEUE[Requeue the NaBSL Reconcile]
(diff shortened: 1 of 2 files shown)


internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -163,6 +164,44 @@ func (r *NonAdminBackupStorageLocationReconciler) initNaBSLDelete(ctx context.Co
...
+
+		logger.V(1).Info("Deleting NonAdminBackup", "backup", nonAdminBackup.Name)
+
+		if err := r.Delete(ctx, &nonAdminBackup); err != nil {
+			if apierrors.IsNotFound(err) {
+				// Ignore NotFound errors (already deleted)
+				continue
+			}
+			logger.Error(err, "Failed to delete NonAdminBackup", "backup", nonAdminBackup.Name)
+			// Continue deleting other backups, this should not block the deletion of the NonAdminBackupStorageLocation
```
mateusoliveira43, line 195: this comment does not make much sense 

GC will remove orphaned Velero objects in OADP namespace (those non admin can not interact with)

the idea of this PR is to avoid mismatches in NonAdminBackup and Velero Backup

problem: non admin user creates NABSL -> creates NAB -> deletes NABSL -> recreates NABSL. Since BSL created by NABSL will always have a different name, if we do not delete NABSL's NABs when deleting it, non admin user can not use NAB `spec.delete`, because DeleteBackupRequest will reference a BSL that does not exist. To avoid scenarios like this, we are deleting the NABs

So, I think `continue` is not right here

mpryc: So should NABSL controller error on such condition and not remove NABSL finalizer?

mpryc: I think it's ok to continue deleting NABackups here, as we only call delete on those objects. If somethng goes wrong in the deletion then NAB controller will cover that?

mpryc: Changed the logic to not continue.

mateusoliveira43 (dismissed):

docs/design/non_admin_backupstoragelocation.md
```diff
@@ -118,4 +124,5 @@ flowchart TD
 1. User deletes the Non-Admin BSL resource.
 2. Controller deletes the Secret from the OADP namespace based on the Non-Admin BSL UUID.
 3. Controller deletes the Velero BSL resource from the OADP namespace based on the Non-Admin BSL UUID.
-4. Controller removes the finalizer from the Non-Admin BSL resource.
+4. Controller deletes the NonAdminBackups for the NaBSL from the user's namespace based on the Non-Admin BSL UUID.
+5. Controller removes the finalizer from the Non-Admin BSL resource.
```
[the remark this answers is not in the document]
mpryc: @kaovilai 
The Controller removes the finalizer from the Non-Admin BSL resource, this is true.

The responsibility of deletion of Velero Backup object is on the NAB Controller:
https://github.com/migtools/oadp-non-admin/blob/11abd720930b2899b709027eed86773d424c4372/internal/controller/nonadminbackup_controller.go#L114-L117

Diagram for this part is here:
https://github.com/migtools/oadp-non-admin/blob/master/docs/design/nab_and_nar_status_update.md#status-update-scenarios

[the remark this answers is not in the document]
shubham-pampattiwar: > that is missing in this design.
Also we should consider a case where velero backup deletion fails, if so, NAB should not be removed until velero backup deletion succeeds.

For this case we have GC controller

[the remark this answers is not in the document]
mpryc: @kaovilai you are right that diagram should make it more explicit. It's not NaBSL Controller that removes the Non Admin Backup and it's Velero Backups. The NaBSL controller only calls k8s delete() on that object and the entire reposnsibility of that removal is on the NAB (Backup) controller.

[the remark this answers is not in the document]
mpryc: @kaovilai answer to another question of yours: NonAdmin Backup won't be removed if Velero Backup objects are still there. Please see on line 117 function `removeNabFinalizerUponVeleroBackupDeletion`

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -163,6 +164,42 @@ func (r *NonAdminBackupStorageLocationReconciler) initNaBSLDelete(ctx context.Co
...
+	}
+
+	if len(nonAdminBackupList.Items) == 0 {
+		logger.V(1).Info("No NonAdminBackups found in NonAdminBackupStorageLocation namespace", "nabsl", nabsl.Name)
+		return false, nil
+	}
+
+	for _, nonAdminBackup := range nonAdminBackupList.Items {
+		// Ensure it belongs to this StorageLocation
+		if nonAdminBackup.Spec.BackupSpec == nil || nonAdminBackup.Spec.BackupSpec.StorageLocation != nabsl.Name {
```
kaovilai, line 183: Ideally this check should be nabsl.uuid

kaovilai: Otherwise this code will act on failed to delete NABs from following
> [When the delete() call on the Non Admin Backup object from the NaBSL controller succeded, however during the deletion of Velero Backup triggered by the delete() call fails for some reason, then the NaBSL object will get removed, but the Non Admin Backup object will not get finalizer removal and will not be deleted](https://github.com/migtools/oadp-non-admin/pull/163/files#r1937852423)

docs/design/non_admin_backupstoragelocation.md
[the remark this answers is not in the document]
mpryc: @kaovilai updated design and diagram. Added to the diagram new subgraph to epmasize it's NAB Controller which is responsible for removing of the NAB objects. NABSL controller only calls delete() on those.

internal/controller/nonadminbackupstoragelocation_controller.go
kaovilai: NABs should have annotations of its NaBSL's UUID which you can reference. Don't add UUID to NAB spec.

kaovilai: And you can use NaBSL kube object UUID for this, shouldn't need to add a new uuid field to nabsl spec either.

mpryc: NaBSL is the object created by the user and not the NaBSL controller, so I am not sure why we would want to generate UUID for it? It will be unique name as the user can not have two same objects within the same namespace.

The UUIDs are important to make relation in the oadp namespace between created by the controller Velero Backup and Velero BSL objects.

mpryc: We store the NaBSL UUID as part of the NAB Status, so user can not change it, we could use it, but as above I am not sure if it will give us anything?

https://github.com/migtools/oadp-non-admin/blob/11abd720930b2899b709027eed86773d424c4372/internal/controller/nonadminbackupstoragelocation_controller.go#L315-L320

Similar way of storing UUIDs is for Non Admin Backup and Non Admin Restore

mpryc: > And you can use NaBSL kube object UUID for this, shouldn't need to add a new uuid field to nabsl spec either.

You are (probably) right. This is mostly to be similar across all NAC objects. For Velero Backup we actually need other then UUID from cluster, because when Velero Sync Controller recreates that Velero Backup object (e.g. on a new cluster) the UUID is different and we can not map it back to NaBSL objects which created them.

kaovilai: I guess NaBSL.status.uuid is already a thing..

kaovilai: > why we would want to generate UUID for it?

I never said that.

mpryc: > I guess NaBSL.status.uuid is already a thing..

We generate this on a first NaBSL creation and then we reference it in the Velero BSL. Do you think it's wrong?

kaovilai: > It will be unique name as the user can not have two same objects within the same namespace.

User can delete NaBSL and recreate. Without checking for matching NaBSL uuid on the NAB annotations, you can potentially be deleting NAB on NaBSLCreatedInMarch when NAC should only delete NAB for NaBSLCreatedInApril.

kaovilai: > We generate this on a first NaBSL creation and then we reference it in the Velero BSL. Do you think it's wrong?
For UUID meant for avoiding collisions in the same cluster, I believe kube provided is sufficient and object.status.uuid is redundant.

If we have cross cluster logic to avoid collision of NaBSL then yes NaBSL.status.UUID could proved useful. Then there needs to be mechanics to upload UUID to the bucket IMO to "save" the UUID for the bucket.

mpryc: I think this will not happen. The user deletes NaBSL based on name, so user is always deleting one object. The controller is removing all the dependencies based on generated UUIDs, which for both user objects are different as we don't generate UUID based on name/namespace, but we actually make it unique and guarantee they will never be same (collision probability is extremely minimal - UUID4 with additional prefix)

kaovilai: > The controller is removing all the dependencies based on generated UUIDs

only if NAC is running.

kaovilai: Many kube users overcommit their clusters, there is always a chance a controller is not running.

mpryc: Ah right so you are suggesting to include UUID information from NaBSL inside the NAB object and also ensure they do match. Got it! In normal operation this should not happen as the NaBSL removal should remove Backups when NAC is running, but I see your point ! Think this would be follow up work ?

kaovilai: Yes.

kaovilai (approved):
LGTM!

mateusoliveira43 (approved):

shubham-pampattiwar (approved):

Issue #141, opened 2025-01-22
Failed NAB reconciles again after DPA update


stillalearner:
Description:

When I created a NAB CR in a non-admin namespace, with some enforcedBackupSpec value mismatch to what is set already in DPA by admin (for example snapshotMoveData), NAB validation fails at first, But on updating the DPA  enforcedBackupSpec value to match the value with the NAB backup spec, It reconciles and start running again.

Steps to reproduce:

1. Create a DPA with Non ADmin enabled with some enforcedBackupSpec value set , in this case snapshotMoveData : true

DPA:

`apiVersion: v1
items:
- apiVersion: oadp.openshift.io/v1alpha1
  kind: DataProtectionApplication
  metadata:
    creationTimestamp: "2025-01-22T11:52:33Z"
    generation: 15
    name: ts-dpa
    namespace: openshift-adp
    resourceVersion: "306326"
    uid: f8d32ddf-c943-4bc5-a6a9-5b4b75454723
  spec:
    backupLocations:
    - velero:
        credential:
          key: cloud
          name: cloud-credentials-gcp
        default: true
        objectStorage:
          bucket: oadp108491vsrf7
          prefix: velero
        provider: gcp
    configuration:
      nodeAgent:
        enable: true
        uploaderType: kopia
      velero:
        defaultPlugins:
        - csi
        - gcp
        - openshift
    nonAdmin:
      enable: true
      enforceBackupSpec:
        snapshotMoveData: true
    unsupportedOverrides:
      tech-preview-ack: "true"
  status:
    conditions:
    - lastTransitionTime: "2025-01-22T11:52:33Z"
      message: Reconcile complete
      reason: Complete
      status: "True"
      type: Reconciled
`

2. Go to Non admin namespace as non admin user, and create a Non Admin Backup CR (lets say backuptest1)

`$ oc get nab backuptest1 -o yaml
apiVersion: oadp.openshift.io/v1alpha1
kind: NonAdminBackup
metadata:
  creationTimestamp: "2025-01-22T15:23:12Z"
  finalizers:
  - nonadminbackup.oadp.openshift.io/finalizer
  generation: 1
  labels:
    app.kubernetes.io/created-by: oadp-operator
    app.kubernetes.io/instance: nonadminbackup-sample
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: nonadminbackup
    app.kubernetes.io/part-of: oadp-operator
  name: backuptest1
  namespace: sns-3
[...]

PR #171, opened 2025-02-06
Improved error messages for enforced backup/restore spec

sseago:
Enforced backup/restore spec validation errors now include the enforced value in addition to the field name.

## Why the changes were made
Enforce spec fields for NA backup and restore listed the invalid field but not the enforced value. This PR updates the error message to display the expected value.

Fixes https://github.com/migtools/oadp-non-admin/issues/141
Update the DPA to specify values for enforceBackupSpec and enforceRestoreSpec -- something like the below example, although you could use different fields to test with
```
    nonAdmin:
      enable: true
      enforceBackupSpec:
        snapshotMoveData: true
      enforceRestoreSpec:
        restorePVs: false
```
Create a NAB which violates the above. For example, with `spec.backupSpec.snapshotMoveData=false`. Expected error message will now be:
```
The administrator has restricted spec.backupSpec.snapshotMoveData field value to true.
```
Create a NAr which violates the above. For example, with `spec.restoreSpec.restorePVs=true`. Expected error message will now be:
```
The administrator has restricted spec.restoreSpec.restorePVs field value to false.
```


weshayutin (dismissed):
thank you @sseago!!

kaovilai (approved):

kaovilai (approved):

shubham-pampattiwar (approved):

mateusoliveira43 (commented):
/lgtm

Issue #144, opened 2025-01-22
NaBSL prefix needs to be unique


shubham-pampattiwar:
We need a unique prefix per NaBSL object so as to avoid scenarios for proper functioning of NaSYNC controller. This would ensure that every NaBSL has a unique S3/object storage URL.

mpryc:
We need to think about scenario where system administrator would want to enforce prefix.

I think that NAC prefix should second, just after what user or system/administrator specifies. If nothing is being specified by the user or enforced by the administrator then it's just NAC prefix. This way we will allow nicely to see them in the s3 storage.

PR #172, opened 2025-02-06
Setting NaBSL prefix 

shubham-pampattiwar:
## Why the changes were made

Fixes https://github.com/migtools/oadp-non-admin/issues/144

- Try creating a NaBSL object by setting a prefix 
- Try creating a NaBSL object without setting a prefix
- Run unit tests (`make simuation-test`)

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -632,3 +632,13 @@ func checkLabelAnnotationValueIsValid(labelsOrAnnotations map[string]string, key
 func GetLogger(ctx context.Context, obj client.Object, key string) logr.Logger {
 	return log.FromContext(ctx).WithValues(key, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()})
 }
+
+// ComputePrefixForObjectStorage returns the prefix to be used for the BackupStorageLocation.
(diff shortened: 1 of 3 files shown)


kaovilai (commented):
Is there a uniqueness expected here of the prefix? Is it on the user?

mateusoliveira43:
@tiger yes, but not on user side, on our side, to avoid synchronizing wrong backups

mpryc:
> > Tested couple of scenarios and only question which I have is around user specyfing '/' or '//' as a prefix. This would translate to an empty string folder which contains another empty string folder which is not allowed in AWS s3.
> > We may consider further validation of the prefix, however ideally if it was directly in the velero BSL validation not NABSL responsibility ?
> 
> yeah double slash not allowed ref: https://www.ietf.org/rfc/rfc2396.txt
> 
> Interim fix : add a NABSL Validation Long term: Add a validation at Velero BSL level We should do both IMO, I will update the PR and create an upstream issue as well. Does that sound good ? @mpryc

@shubham-pampattiwar Yes, can be of course separate PR to make it moving. This one works for me and does what it tells to do.

/lgtm

mpryc:
@shubham-pampattiwar added hold to make it your call if you want to update this PR or create new one. I am fine either way.

PR #173, opened 2025-02-07
Ensure instructions to build work on arm and amd

weshayutin:
w/o using dockerx builds the normal docker-build, docker-push failed to work properly as the setup documentation stated on ARM
clusters.   This change simply makes the build work more like
oadp-operator.

## Why the changes were made

building w/ ARM failed when using
https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md

Tested w/ both ARM and AMD clusters.  :)  works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,13 @@ GOBIN=$(shell go env GOPATH)/bin
 else
 GOBIN=$(shell go env GOBIN)
 endif
+OC_CLI ?= $(shell which oc)
+TARGETARCH = $(shell $(OC_CLI) get node -o jsonpath='{.items[0].status.nodeInfo.architecture}' 2> /dev/null)
+
+
+# .PHONY: print-arch
+# print-arch: ## Print the target architecture
+# 	@echo $(TARGETARCH)
 
 # CONTAINER_TOOL defines the container tool to be used for building images.
 # Be aware that the target commands are only tested with Docker which is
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,6 +11,13 @@ GOBIN=$(shell go env GOPATH)/bin
 else
 GOBIN=$(shell go env GOBIN)
 endif
+OC_CLI ?= $(shell which oc)
+TARGETARCH = $(shell $(OC_CLI) get node -o jsonpath='{.items[0].status.nodeInfo.architecture}' 2> /dev/null)
+
+
+# .PHONY: print-arch
```
mateusoliveira43, line 18: delete leftover test

kaovilai:
Thoughts on just making multi platform images always? This controller image should be small enough that its going to create productivity not having to ask someone to rebuild for their arch

weshayutin:
> Thoughts on just making multi platform images always? This controller image should be small enough that its going to create productivity not having to ask someone to rebuild for their arch for less than a minute of additional build time.

That is a valid point and one side of the issue.  If I recall correctly folks involved in this project perfered to build quickly vs build all arch types ( it was annoying ).  

We def do not need to build for ppc or s390.  

IIUC, we are building oadp-operator based on the arch of the cluster.  Seems like the right pattern to follow here as well.

Makefile
weshayutin: yup.. fixed.  Thank you

mpryc:
@weshayutin which way will this PR go? multi-arch as @kaovilai suggested or rebase to resolve Makefile conflict?

kaovilai:
The operator takes a long time hence only single arch. Also it would've required a lot of rework of oadp makefile to build mutiarch oadp.

This repo is not the case because of lack of operator ask bundle building

PR #182, opened 2025-02-07
fix(Makefile): docker-buildx-nac do not add duplicate `--platform`

kaovilai:
Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

## Why the changes were made

Same motivation as https://github.com/kubernetes-sigs/kubebuilder/pull/4545

run `make docker-buildx-nac` in this repo


Makefile
kaovilai, line 1: While it could be argued that this target does not work with podman, it's from kubebuilder template which also use the same $(CONTAINER_TOOL).

And I am PR'ing into kubebuilder in https://github.com/kubernetes-sigs/kubebuilder/pull/4545

mpryc: See https://github.com/kubernetes-sigs/kubebuilder/issues/4544#issuecomment-2643500648

mpryc: @kaovilai it's your call if we want to add it or not, but it should be commented that it differs from upstream kubebuilder, so we don't simply replace it in the future by following for example instructions like:
https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.34.0/

e.g. Without clear comment to the developer I would replace everything within `docker-buildix` including diff from upstream kubebuilder.

mpryc: @kaovilai could you please add the comment to the code and I will approve?

kaovilai: or do you prefer a third target that would not conflict with kubebuilder scaffold?

mpryc: @kaovilai that is cool idea (third target) - I like it more then comment.

Issue #189, opened 2025-02-11
Exclude NAC resources (NAB, NAR, NABSL) from Non-Admin backups


shubham-pampattiwar:
We should exclude NAC resources from Non-Admin backups via excludedResources spec even if the user does not explicitly specify them

PR #193, opened 2025-02-16
Exclude NAC resources from backup

shubham-pampattiwar:
## Why the changes were made

Fixes https://github.com/migtools/oadp-non-admin/issues/189

- Try creating a NAB backup
- Check the backup spec and resource List as well once the backup completes
- At both the places, NAC resources must be absent

### DIFF
--- a/internal/controller/nonadminbackup_controller.go
+++ b/internal/controller/nonadminbackup_controller.go
@@ -641,6 +641,13 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
 			backupSpec.StorageLocation = nonAdminBsl.Status.VeleroBackupStorageLocation.Name
 		}
 
+		// Exclude NAC resources (NAB, NAR, NABSL) from Non-Admin backups
+		backupSpec.ExcludedResources = []string{
+			nacv1alpha1.NonAdminBackups,
+			nacv1alpha1.NonAdminRestores,
+			nacv1alpha1.NonAdminBackupStorageLocations,
+		}
+
 		veleroBackup := velerov1.Backup{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:        veleroBackupNACUUID,
--- a/api/v1alpha1/nonadmin_types.go
+++ b/api/v1alpha1/nonadmin_types.go
@@ -50,3 +50,9 @@ type QueueInfo struct {
 	// estimatedQueuePosition is the number of operations ahead in the queue (0 if not queued)
 	EstimatedQueuePosition int `json:"estimatedQueuePosition"`
 }
+
+const (
+	NonAdminBackups                = "nonadminbackups"
+	NonAdminRestores               = "nonadminrestores"
+	NonAdminBackupStorageLocations = "nonadminbackupstoragelocations"
+)
--- a/internal/controller/nonadminbackup_controller_test.go
+++ b/internal/controller/nonadminbackup_controller_test.go
@@ -1086,6 +1086,11 @@ var _ = ginkgo.Describe("Test full reconcile loop of NonAdminBackup Controller",
 					ginkgo.By("Validating Velero Backup Spec")
 					expectedSpec := scenario.enforcedBackupSpec.DeepCopy()
 					expectedSpec.IncludedNamespaces = []string{nonAdminObjectNamespace}
+					expectedSpec.ExcludedResources = []string{
+						nacv1alpha1.NonAdminBackups,
+						nacv1alpha1.NonAdminRestores,
+						nacv1alpha1.NonAdminBackupStorageLocations,
+					}
 					gomega.Expect(reflect.DeepEqual(veleroBackup.Spec, *expectedSpec)).To(gomega.BeTrue())
 				}
 


internal/controller/nonadminbackup_controller.go
```diff
@@ -641,6 +641,13 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
 			backupSpec.StorageLocation = nonAdminBsl.Status.VeleroBackupStorageLocation.Name
 		}
 
+		// Exclude NAC resources (NAB, NAR, NABSL) from Non-Admin backups
```
mateusoliveira43, line 644: how is this done in velero side? can a velero backup include a velero backup CR?

There is a velero plugin PR as well for this. Is this PR necessary?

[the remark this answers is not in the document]
mpryc: Do I understand this comment correctly?
If the user exclude some resources, this change will ignore what user excluded and only exclude the ones which are defined below, so what we want is?

```go
backupSpec.ExcludedResources = append(backupSpec.ExcludedResources, 
    nacv1alpha1.NonAdminBackups,
    nacv1alpha1.NonAdminRestores,
    nacv1alpha1.NonAdminBackupStorageLocations,
)
```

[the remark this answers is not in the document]
sseago: @mpryc @mateusoliveira43 I think you're right. What we want to do is add to the exclude list. So if user excludes secrets, then we'd exclude secrets *and* NAB, NAR, etc.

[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the PR, Thanks for the catch folks !

shubham-pampattiwar: > how is this done in velero side? can a velero backup include a velero backup CR?
I am not sure velero excludes them, we need to exclude NACs so that we have proper functioning of our inter-ns sync.
> There is a velero plugin PR as well for this. Is this PR necessary?
The plugin PR is to skip restore of resources using RIA (RestoreItemAction)
This PR is to exclude them from backups. The plugin PR serves as an additional check.

sseago: @mateusoliveira43 it's not really an issue on velero side since we don't support backing up the velero NS. It's a problem for NAC resources because those are located in the user workload NS.

internal/controller/nonadminbackup_controller.go
```diff
@@ -641,6 +641,12 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
...
+		backupSpec.ExcludedResources = append(backupSpec.ExcludedResources,
+			nacv1alpha1.NonAdminBackups,
+			nacv1alpha1.NonAdminRestores,
+			nacv1alpha1.NonAdminBackupStorageLocations)
+
```
sseago, line 649: I just realized that this will break velero validation if the NAB spec sets a value for ExcludedNamespacedResources -- since there's velero validation to ensure that we don't include both old and new style includes/excludes. I think you need to look at the current value for backupSpec.ExcludedResources and backupSpec.ExcludedNamespaceScopedResources and append these by whichever is nonempty. Actually, you have to look at the full list of params. if all new-style params are empty, then append to ExcludedResources, else append to ExcludedNamespaceScopedResources.

mateusoliveira43: good catch

Ref https://github.com/vmware-tanzu/velero/blob/80cea31a8457835086371d9036f67ecb150208b6/pkg/controller/backup_controller.go#L460-L467

https://github.com/vmware-tanzu/velero/blob/80cea31a8457835086371d9036f67ecb150208b6/pkg/controller/backup_controller.go#L922-L932

internal/controller/nonadminbackup_controller.go
mateusoliveira43: right, thanks for explanations 

With this, NAB wont include NAC CRs, but standalone backups will. Could customers complain that it see a resource in backup but it is not being restored? Question here is if it would be better to add exclusion more higher level (like in BIA, do not know if it is possible)

internal/controller/nonadminbackup_controller.go
mpryc: Maybe in the NAB we should *not* allow to use old filter parameters on the NAB validation part?

mateusoliveira43: I agree we should prioritize using new syntax, but since old syntax it is still allowed in velero, I think is ok to use it here

sseago: From what I've seen, the vast majority of users still use the old syntax. Also, it's not deprecated, so it should be fully supported.

sseago: It does complicate the logic here, though -- since we have to determine which syntax the user is using based on several spec fields and append to the appropriate field.

sseago: Basically, the logic here would be
```
if haveNewResourceFilterParameters {
    append(backupSpec.ExcludedNamespaceScopedResources...
} else {
    append(backupSpec.ExcludedResources...
}
```
where `haveNewResourceFilterParameters` is defined as it is here:
https://github.com/vmware-tanzu/velero/blob/80cea31a8457835086371d9036f67ecb150208b6/pkg/controller/backup_controller.go#L926-L929

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: @mateusoliveira43 We can document that NAC resources are explicitly excluded in NAB spec.
@sseago IIRC we used add some resources in excluded List for MTC too.

internal/controller/nonadminbackup_controller.go
shubham-pampattiwar: Thanks you all ! updated the PR, PTAL !

mpryc (commented):
/lgtm

internal/controller/nonadminbackup_controller.go
sseago: @mateusoliveira43 BIA doesn't currently allow discarding the resource but RIA does. The reason for that is that when I first implemented the discarding feature, we didn't have plugin versioning -- RIA API has response in a single struct, while BIA returns several separate params. Adding a field to the struct could be done without breaking the API, while on the BIA side, adding a return param would break existing plugins. We could at this point add the field with a BIA version bump, although it's a relatively low priority. Discarding on restore has the same effect as discarding on backup -- the only downside would be if you're discarding something that is used in large quantities, it could have a performance/storage impact. For something like NAC resources, though, that's not a concern. It may be worth documenting that if OADP admin backs up namespaces with NAC resources, the backup list will include NAC CRs but they won't be restored.

Issue #201, opened 2025-02-18
Handle `includeClusterResources` and `includeClusterScopedResources` for NAB spec


shubham-pampattiwar:
For `includeClusterResources`, the acceptable values for non-admin users are `nil` and `false`. Anything else should fail the NAB operation in `BackingOff` phase
Similarly, for `IncludedClusterScopedResources`, the acceptable value is an empty list.

PR #208, opened 2025-02-19
Add validation error on NAB for restricted spec values

shubham-pampattiwar:
## Why the changes were made

Fixes https://github.com/migtools/oadp-non-admin/issues/200
Fixes https://github.com/migtools/oadp-non-admin/issues/201

Related to https://github.com/openshift/oadp-operator/pull/1632

- Try specifying excludedNamespaces, IncludeClusterResources as true, and non-empty IncludedClusterScopedResources  on NAB backup spec
- NAB should error out

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -100,10 +100,12 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 		return fmt.Errorf("NonAdminBackup spec.backupSpec.excludedNamespaces is restricted")
 	}
 
-	if enforcedBackupSpec.IncludedNamespaces != nil {
-		if !containsOnlyNamespace(enforcedBackupSpec.IncludedNamespaces, nonAdminBackup.Namespace) {
-			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces enforced value by admin user violates NAC usage")
-		}
+	if nonAdminBackup.Spec.BackupSpec.IncludeClusterResources != nil && *nonAdminBackup.Spec.BackupSpec.IncludeClusterResources {
+		return fmt.Errorf("NonAdminBackup spec.backupSpec.includeClusterResources can not be set or must be set to false")
+	}
+
+	if len(nonAdminBackup.Spec.BackupSpec.IncludedClusterScopedResources) > 0 {
+		return fmt.Errorf("NonAdminBackup spec.backupSpec.includedClusterScopedResources is restricted, only an empty list is allowed")
 	}
 
 	if nonAdminBackup.Spec.BackupSpec.StorageLocation != constant.EmptyString {
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -92,20 +92,20 @@ func containsOnlyNamespace(namespaces []string, namespace string) bool {
 func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpNamespace string, nonAdminBackup *nacv1alpha1.NonAdminBackup, enforcedBackupSpec *velerov1.BackupSpec) error {
 	if nonAdminBackup.Spec.BackupSpec.IncludedNamespaces != nil {
 		if !containsOnlyNamespace(nonAdminBackup.Spec.BackupSpec.IncludedNamespaces, nonAdminBackup.Namespace) {
-			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces can not contain namespaces other than: %s", nonAdminBackup.Namespace)
+			return fmt.Errorf(constant.NABRestrictedErr+", can not contain namespaces other than: %s", "spec.backupSpec.includedNamespaces", nonAdminBackup.Namespace)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


internal/common/function/function.go
```diff
@@ -95,6 +95,11 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces can not contain namespaces other than: %s", nonAdminBackup.Namespace)
 		}
 	}
+
+	if nonAdminBackup.Spec.BackupSpec.ExcludedNamespaces != nil {
```
shubham-pampattiwar, line 99: This also means that the admins cannot enforce this spec value.

mateusoliveira43: what you mean by that?

we need code to check `enforcedBackupSpec` to do not allow admins enforcing it

shubham-pampattiwar: operator PR https://github.com/openshift/oadp-operator/pull/1632

kaovilai: do we want to clarify which error can be lifted by admin to nonadmin users?

internal/common/function/function.go
```diff
@@ -95,10 +95,17 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
...
+		return fmt.Errorf("NonAdminBackup spec.backupSpec.excludedNamespaces is restricted")
+	}
+
+	if nonAdminBackup.Spec.BackupSpec.IncludeClusterResources != nil && *nonAdminBackup.Spec.BackupSpec.IncludeClusterResources {
+		return fmt.Errorf("NonAdminBackup spec.backupSpec.includeClusterResources can not be set or must be set to false")
+	}
+
+	if len(nonAdminBackup.Spec.BackupSpec.IncludedClusterScopedResources) > 0 {
+		return fmt.Errorf("NonAdminBackup spec.backupSpec.includedClusterScopedResources is restricted, only an empty list is allowed")
```
mpryc, line 108: Please make this error more informative. What happens when empty list is entered?

Similar error to the enforced admin?
```go
return false, fmt.Errorf("DPA spec.nonAdmin.includedClusterScopedResources is restricted and must remain empty; it cannot be specified")
```

internal/common/function/function_test.go
```diff
@@ -409,23 +419,6 @@ func TestValidateBackupSpecEnforcedFields(t *testing.T) {
 			}
 		})
 	}
-
```
mateusoliveira43, line 412: I do not like this change

idea of this is to help us test if new fields are added to velero API

I would `t.Run` inside `for` to run different test for restricted fields

Example 
```go
            if slices.Contains([]string{"ExcludedNamespaces"}, test.name) {
				reflect.ValueOf(userNonAdminBackup.Spec.BackupSpec).Elem().FieldByName(test.name).Set(reflect.ValueOf(test.overrideValue))
				err := ValidateBackupSpec(context.Background(), fakeClient, "oadp-namespace", userNonAdminBackup, enforcedSpec)
				if err == nil {
					t.Errorf("setting restricted backup spec field '%v' test failed: %v", test.name, err)
				}
				return
			}
```

shubham-pampattiwar: We are not referring the actual Velero API in the tests here. So in a case when Velero API get updated our unit tests won't automatically start failing, right ? We will still need to add tests for manually testing spec values that are enforceable.  Please correct me if I am wrong here.
In the current PR we are testing values that are enforceable in `TestValidateBackupSpecEnforcedFields` and those which have validation error are tested under `TestValidateBackupSpec`. I believe this much is adequate here.

shubham-pampattiwar: Done, updated the PR !

weshayutin (approved):

internal/common/function/function.go
```diff
@@ -92,13 +92,20 @@ func containsOnlyNamespace(namespaces []string, namespace string) bool {
...
-		if !containsOnlyNamespace(enforcedBackupSpec.IncludedNamespaces, nonAdminBackup.Namespace) {
-			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces enforced value by admin user violates NAC usage")
-		}
+
+	if nonAdminBackup.Spec.BackupSpec.ExcludedNamespaces != nil {
+		return fmt.Errorf(constant.NABRestrictedErr, "spec.backupSpec.excludedNamespaces")
+	}
+
+	if nonAdminBackup.Spec.BackupSpec.IncludeClusterResources != nil && *nonAdminBackup.Spec.BackupSpec.IncludeClusterResources {
+		return fmt.Errorf(constant.NABRestrictedErr+", can only be set to false", "spec.backupSpec.includeClusterResources")
```
sseago, line 104: NIT: nil and false are both allowed, so this error message might confuse someone if they set it to true, causing them to then set it to false -- which will pass validation but won't work with PVs.

sseago (approved):
Approving so we can get this in for testing now, but the error message I noted should probably be fixed in a follow-on PR.

PR #209, opened 2025-02-19
Add Validation error for includeClusterResources and includedClusterScopedResources

shubham-pampattiwar:
## Why the changes were made

https://github.com/migtools/oadp-non-admin/issues/201

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -95,6 +95,15 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces can not contain namespaces other than: %s", nonAdminBackup.Namespace)
 		}
 	}
+
+	if nonAdminBackup.Spec.BackupSpec.IncludeClusterResources != nil && *nonAdminBackup.Spec.BackupSpec.IncludeClusterResources {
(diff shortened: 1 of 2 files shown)


internal/common/function/function.go
```diff
@@ -95,6 +95,15 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 			return fmt.Errorf("NonAdminBackup spec.backupSpec.includedNamespaces can not contain namespaces other than: %s", nonAdminBackup.Namespace)
 		}
 	}
+
+	if nonAdminBackup.Spec.BackupSpec.IncludeClusterResources != nil && *nonAdminBackup.Spec.BackupSpec.IncludeClusterResources {
+		return fmt.Errorf("Non-admin users are not allowed to set includeClusterResources to true, it must be either nil or false")
```
mateusoliveira43, line 100: would change message here to be more user friendly (it may not know what `nil` means)

and would also be more precise: **includeClusterResources** -> **NonAdminBackup spec.backupSpec.includeClusterResources**

what about something like this
```
NonAdminBackup spec.backupSpec.includeClusterResources can not be set or must be set to false
```

shubham-pampattiwar:
Closing this PR, will combine the changes from here in https://github.com/migtools/oadp-non-admin/pull/208

Issue #195, opened 2025-02-17
NonAdminBackup status should clearly mention that the associated velero backup has been removed


PrasadJoshi12:
When velero backup is deleted/expired and removed by velero controller, The associated nonAdminBackup should indicate that clearly for nonAdmin user. Currently it shows the backup is getting delete but in reality backup can be gone since long time.

Steps to follow:-

1. Create  a non admin backup with smaller ttl or remove backup after it gets completed
```console
$ cat nab.yaml 
apiVersion: oadp.openshift.io/v1alpha1
kind: NonAdminBackup
metadata:
  name: test-nab
spec:
  backupSpec:
    includedNamespaces:
    - "test-nac"
    ttl: 0h01m0s
```

The CR shows the velero backup status as deleting. 

```yaml
veleroBackup:
    nacuuid: test-nac-test-nab1-a7b201d5-7663-4ab3-a654-eeca67aebfcc
    name: test-nac-test-nab1-a7b201d5-7663-4ab3-a654-eeca67aebfcc
    namespace: openshift-adp
    status:
      completionTimestamp: "2025-02-17T12:17:14Z"
      expiration: "2025-02-17T12:18:10Z"
      formatVersion: 1.1.0
      hookStatus: {}
      phase: Deleting
      progress:
        itemsBackedUp: 21
        totalItems: 21
      startTimestamp: "2025-02-17T12:17:10Z"
      version: 1
```

shubham-pampattiwar:
@mpryc @mateusoliveira43  In this scenario, won't the NAB object get deleted as well ?

mateusoliveira43:
no, because we do not reconcile on velero backup deletion (predicate side)

but should be easy fix

I would go with just status update, even though some special cases (like backup expired) would make sense to delete NAB as well

shubham-pampattiwar:
Hmm, maybe status update as well as a message on how to cleanup NAB ?

Issue #178, opened 2025-02-07
NAR CR goes into BackingOff phase when associated backup is removed


PrasadJoshi12:
Steps followed:-
Create Non Admin Backup
Created NonAdminRestore. (wait for restore to complete)

```console
$ oc get nar -n test -o yaml
apiVersion: v1
items:
- apiVersion: oadp.openshift.io/v1alpha1
  kind: NonAdminRestore
  metadata:
    creationTimestamp: "2025-02-06T11:01:02Z"
    finalizers:
    - nonadminrestore.oadp.openshift.io/finalizer
    generation: 2
    name: test-nar
    namespace: test
    resourceVersion: "126191"
    uid: eceae35a-e6a9-4a95-855a-073c2c8132e8
  spec:
    restoreSpec:
      backupName: test
      hooks: {}
      itemOperationTimeout: 0s
  status:
    conditions:
    - lastTransitionTime: "2025-02-06T11:01:02Z"
      message: restore accepted
      reason: RestoreAccepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2025-02-06T11:01:02Z"
      message: Created Velero Restore object
      reason: RestoreScheduled
      status: "True"
      type: Queued
    phase: Created
    queueInfo:
      estimatedQueuePosition: 0
    veleroRestore:
      nacuuid: test-test-nar-465e530d-2de7-4598-94ca-4d85d9f3f892
      name: test-test-nar-465e530d-2de7-4598-94ca-4d85d9f3f892
      namespace: openshift-adp
      status:
        completionTimestamp: "2025-02-06T11:01:04Z"
        hookStatus: {}
        phase: Completed
        progress:
          itemsRestored: 15
          totalItems: 15
        startTimestamp: "2025-02-06T11:01:02Z"
        warnings: 4
```
3. Now delete NonAdminBackup CR by setting deleteBackup: true
4. Observe NAR CR
```yaml
apiVersion: oadp.openshift.io/v1alpha1
  kind: NonAdminRestore
  metadata:
    creationTimestamp: "2025-02-06T11:01:02Z"
    finalizers:
    - nonadminrestore.oadp.openshift.io/finalizer
    generation: 2
    name: test-nar
    namespace: test
    resourceVersion: "126418"
    uid: eceae35a-e6a9-4a95-855a-073c2c8132e8
  spec:
    restoreSpec:
      backupName: test
      hooks: {}
      itemOperationTimeout: 0s
  status:
    conditions:
[...]

PR #210, opened 2025-02-19
fix: NonAdmin objects associated objects deletion info

mateusoliveira43:
## Why the changes were made

Fix https://github.com/migtools/oadp-non-admin/issues/178

Fix https://github.com/migtools/oadp-non-admin/issues/195

Also add checks to avoid non admin users to maliciously creating NABs that violates NAC rules or using backups from other non admin users.

~Blocked by https://github.com/openshift/oadp-operator/pull/1640~

Follow [install-from-source](https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#install-from-source) testing documentation, pointing this branch and my OADP fork branch [`fix/nac-associated-objects-deletion`](https://github.com/mateusoliveira43/oadp-operator/tree/fix/nac-associated-objects-deletion)

Create NAB, create NAR and then delete associated velero restore. It should be clear in NAR status that its associated velero restore was deleted and not run another restore.

Create NAB, create NAR and then delete NAB using NAB `spec.deleteBackup`. NAR and associated velero restore should be deleted.

Create NAB and then delete associated velero backup. It should be clear in NAB status that its associated velero backup was deleted and not run another backup.

### DIFF
--- a/internal/controller/nonadminbackup_controller.go
+++ b/internal/controller/nonadminbackup_controller.go
@@ -116,8 +116,8 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		logger.V(1).Info("Executing standard delete path")
 		reconcileSteps = []nonAdminBackupReconcileStepFunction{
 			r.setStatusAndConditionForDeletionAndCallDelete,
+			r.deleteNonAdminRestores,
 			r.createVeleroDeleteBackupRequest,
-			r.removeNabFinalizerUponVeleroBackupDeletion,
 		}
 
 	case !nab.ObjectMeta.DeletionTimestamp.IsZero():
@@ -129,9 +129,8 @@ func (r *NonAdminBackupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		logger.V(1).Info("Executing direct deletion path")
 		reconcileSteps = []nonAdminBackupReconcileStepFunction{
 			r.setStatusForDirectKubernetesAPIDeletion,
-			r.deleteVeleroBackupObjects,
 			r.deleteDeleteBackupRequestObjects,
-			r.removeNabFinalizerUponVeleroBackupDeletion,
+			r.deleteVeleroBackupObjects,
 		}
 
 	case function.CheckLabelAnnotationValueIsValid(nab.Labels, constant.NabSyncLabel):
@@ -196,7 +195,6 @@ func (r *NonAdminBackupReconciler) setStatusAndConditionForDeletionAndCallDelete
 			return false, err
 		}
 		logger.V(1).Info("NonAdminBackup status marked for deletion")
-		requeueRequired = true // Requeue to ensure latest NAB object in the next reconciliation steps
 	} else {
 		logger.V(1).Info("NonAdminBackup status unchanged during deletion")
 	}
@@ -248,6 +246,27 @@ func (r *NonAdminBackupReconciler) setStatusForDirectKubernetesAPIDeletion(ctx c
 	return false, nil
 }
 
+func (r *NonAdminBackupReconciler) deleteNonAdminRestores(ctx context.Context, logger logr.Logger, nab *nacv1alpha1.NonAdminBackup) (bool, error) {
+	logger.V(1).Info("checking for NonAdminRestores to delete")
+	nonAdminRestores := &nacv1alpha1.NonAdminRestoreList{}
+	if err := r.List(ctx, nonAdminRestores, client.InNamespace(nab.Namespace)); err != nil {
+		logger.Error(err, "Failed to list NonAdminRestores in NonAdminBackup namespace")
+		return false, err
+	}
+
+	for _, nonAdminRestore := range nonAdminRestores.Items {
+		if nonAdminRestore.Spec.RestoreSpec.BackupName == nab.Name {
+			if err := r.Delete(ctx, &nonAdminRestore); err != nil {
+				logger.Error(err, "Failed to delete NonAdminRestore in NonAdminBackup namespace")
+				return false, err
+			}
+			logger.V(1).Info("NonAdminRestore deleted")
+		}
+	}
+
+	return false, nil
+}
+
 // createVeleroDeleteBackupRequest initiates deletion of the associated VeleroBackup object
 // that is referenced by the NACUUID within the NonAdminBackup (NAB) object.
 // This ensures the VeleroBackup is deleted before the NAB object itself is removed.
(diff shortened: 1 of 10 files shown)


internal/controller/nonadminrestore_controller.go
```diff
@@ -333,6 +333,28 @@ func (r *NonAdminRestoreReconciler) createVeleroRestore(ctx context.Context, log
 	}
 
 	if veleroRestore == nil {
```
mateusoliveira43, line 335: is this approach ok? when we see NAR associated velero restore got deleted, we just inform this in status to non admin user

We could also just delete it. I am afraid this could lead to bad UX, where something by mistake deletes velero restore and nona mdin user does not know why its restore was deleted (but that is only a CR, restore operation would not be undone)

internal/controller/nonadminbackup_controller.go
mateusoliveira43, line 1: TODO write tests

internal/controller/nonadminrestore_controller.go
mpryc: What would happen if we re-create Velero Restore? So NAR controller reacting on the Velero Restore objects which has NAC labels and ensures the reconcile of NAC re-creates them when they got deleted?

mateusoliveira43: currently, if related velero restore is deleted, it would get recreated by controller

with this PR, if related velero restore is deleted, and then somehow recreated (with same labels and annotations) NAB status would update to match new velero restore status

shubham-pampattiwar: I think status update on NAB should be fine

internal/controller/nonadminrestore_controller.go
```diff
@@ -333,6 +333,28 @@ func (r *NonAdminRestoreReconciler) createVeleroRestore(ctx context.Context, log
...
+		if nar.Status.Phase == nacv1alpha1.NonAdminPhaseCreated {
```
mateusoliveira43, line 336: TODO test

I think I can not use phase here. `meta.IsStatusConditionTrue(nar.Status.Conditions, string(nacv1alpha1.NonAdminConditionQueued))` would be better

scenario I am afraid

NAR created -> (after restore operation finishes) velero restore deleted -> status is updated telling velero restore and also phase to BackingOff -> user updates NAR spec -> new velero restore is created

internal/controller/nonadminbackup_controller.go
```diff
@@ -628,6 +621,33 @@ func (r *NonAdminBackupReconciler) createVeleroBackupAndSyncWithNonAdminBackup(c
 	}
 
 	if veleroBackup == nil {
+		if function.CheckLabelAnnotationValueIsValid(nab.Labels, constant.NabSyncLabel) || nab.Status.Phase == nacv1alpha1.NonAdminPhaseCreated {
+			if function.CheckLabelAnnotationValueIsValid(nab.Labels, constant.NabSyncLabel) {
+				err = errors.New("related Velero Backup to be synced from does not exist")
+			}
+			if meta.IsStatusConditionTrue(nab.Status.Conditions, string(nacv1alpha1.NonAdminConditionQueued)) {
+				// err = errors.New("related Velero Backup was already created, it will not be recreated. To create new Backup, create new NonAdminBackup")
+				err = errors.New("NonAdminBackup is in final stage and its related Velero Backup was deleted")
```
mateusoliveira43, line 630: any suggestion on how this error should be written?

shubham-pampattiwar: "NonAdminBackup is finalized and its associated Velero Backup has been removed. Please create a new NonAdminBackup to initiate a new backup." ?

mpryc: A quick question—aren't we handling this differently than Velero Backup objects? Why add extra logic to inform the user that the Velero backup was deleted? Normally, Velero doesn’t provide this information, so aren’t we introducing inconsistent behavior between Non-Admin Backup and Velero Backup?

internal/controller/nonadminrestore_controller.go
```diff
@@ -333,6 +300,27 @@ func (r *NonAdminRestoreReconciler) createVeleroRestore(ctx context.Context, log
 	}
 
 	if veleroRestore == nil {
+		if meta.IsStatusConditionTrue(nar.Status.Conditions, string(nacv1alpha1.NonAdminConditionQueued)) {
+			err = errors.New("NonAdminRestore is in final stage and its related Velero Restore was deleted")
```
mpryc, line 304: Same question here - aren’t we introducing inconsistent behavior between Non-Admin Backup and Velero Backup with extra logic that may not be needed?

internal/controller/nonadminbackup_controller.go
mateusoliveira43: this is what was decided as how to handle https://github.com/migtools/oadp-non-admin/issues/195#issuecomment-2666825420

what would be related scenario in Velero case?

mpryc: > this is what was decided as how to handle [#195 (comment)](https://github.com/migtools/oadp-non-admin/issues/195#issuecomment-2666825420)

The question mark on the comment is rather not a decision. @shubham-pampattiwar ?

> what would be related scenario in Velero case?

Deleted Velero Backup object from the cluster. If it exists in the s3 it will be synced back and then NAB will get synced as well, so deleted Velero Backup = NAC Controller deletes NAB

internal/controller/nonadminrestore_controller.go
mateusoliveira43: from meeting today, we will only have status message saying related object was deleted

internal/controller/nonadminbackup_controller.go
mpryc: ok

shubham-pampattiwar (approved):

sseago (approved):

shubham-pampattiwar:
@mpryc PTAL, thanks !

weshayutin:
status:
  conditions:
  - lastTransitionTime: "2025-02-26T18:25:13Z"
    message: NonAdminRestore is finalized and its associated Velero Restore has been
      removed. Please create a new NonAdminRestore to initiate a new Restore
    reason: VeleroRestoreNotFound

nice

weshayutin:
nice.. updated deleteBackup:true
and poof.. nab, nar and velero backup, restore were deleted.  very nice..

weshayutin (approved):
/LGTM

Issue #204, opened 2025-02-18
NABSL spec review on what is admin-enforceable and what is restricted for non-admin users.


shubham-pampattiwar:
We need to go through specific scenarios and decide whats supported and whats not.

weshayutin:
discuss with team, for this release... 
this IS the biggest vulnerability.

Cluster Admin needs to be able to  allow/white list NABSL in the DPA.

shubham-pampattiwar:
Proposed approach to solve this problem:
Create workflow:
- Non-admin NS users create NABSL in their own NS
- NAC controller create an `NABSLApprovalRequest` CR in operator NS (`creationApproved: false`)
- Now Velero BSL creation is pending because `NABSLApprovalRequest` isn't approved yet by the admin, NABSL status would reflect that.
- Admin approves the `NABSLApprovalRequest` by updating the `NABSLApprovalRequest` CR. (`creationApproved: true`)
- NAC controller checks for approval before creating the corresponding VBSL
- Once approved, VBSL is created in operator NS.

Update workflow:
- Non-admin users update an existing NABSL CR
- NAC controller updates an existing `NABSLApprovalRequest` CR in operator NS (`updateApproved: false `)
- Admin approves the `NABSLApprovalRequest` by updating the `NABSLApprovalRequest` CR. (`updateApproved: true `)
- NAC controller checks for approval before updating the corresponding VBSL
- Once approved, VBSL is updated in operator NS, and update approval is toggled back `updateApproved: false `, also `lastApprovedSpec: {} ` is stored for journaling purposes.

Proposed `NABSLApprovalRequest`  CR:
```
apiVersion: oadp.openshift.io/v1alpha1
kind: NABSLApprovalRequest
metadata:
  name: nabsl-hash-name
  namespace: openshift-adp<Operator NS, this is the key here>
spec:
  nabslName: nabsl-name
  nabslNamespace: nac-user-ns
  creationApproved: false  # Tracks approval for creation
  updateApproved: false    # Tracks approval for updates
  lastApprovedSpec: {}  # Stores last approved NABSL spec
```
**Note:** 
- This would be an opt-in feature via DPA, Admins will have to specifically enable this workflow. 
- Also we will need to add this ApprovalRequest CR in our cleanup and GC workflows.
- Every NABSL will have only one ApprovalRequest CR
cc: @mpryc @weshayutin

mpryc:
[...]

PR #216, opened 2025-02-24
Approval of the NonAdminBackupStorageLocation

mpryc:
Implements approvals for the NaBSL objects.

Feature which allow to approve NaBSL by the cluster admin prior to the BSL creation.

Depends-on: https://github.com/openshift/oadp-operator/pull/1633

## Why the changes were made

To implement approvals for the cluster admin as described in the https://github.com/migtools/oadp-non-admin/issues/204

Fixes https://github.com/migtools/oadp-non-admin/issues/220

### Disable the feature
1. Deploy oadp operator with the following PR: https://github.com/openshift/oadp-operator/pull/1633
2. Set the DPA to:
```yaml
  nonAdmin:
    enable: true
    requireAdminApprovalForBSL: false
```
3. Nothing should be changed from the user perspective, on the cluster in the openshift-adp namespaces there will be `NonAdminBackupStorageLocationRequest` objects with approved flag
4. Modifying the NaBSLRequest objects and seting them to anything else then approve will make those reverted back to approve state.

### Enable the feature
1. Deploy oadp operator with the following PR: https://github.com/openshift/oadp-operator/pull/1633
2. Set the DPA to:
```yaml
  nonAdmin:
    enable: true
    requireAdminApprovalForBSL: true
```
3. Creating NaBSL should result in the condition not satisfied and requires admin to approve the NaBSLRequest object
4. Update the corresponding NaBSLRequest and set the:
```yaml
spec:
  approvalDecision: approve
```
5. NaBSL should be created
6. Patch the NaBSLRequest and set the flag to:
```yaml
spec:
  approvalDecision: reject # or pending
```
7. The NaBSL should be updated to reflect the decision, corresponding Velero BSL and Secret in the openshift-adp namespace shoud be removed.

### Update the NaBSL spec
This change disallows to update NaBSL spec as a result test can be performed to update spec that was already created. Changes to the spec will result in the BSL to being removed from the cluster and the user informed.


api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// NonAdminBackupStorageLocationRequestSpec defines the desired state of NonAdminBackupStorageLocationRequest
+type NonAdminBackupStorageLocationRequestSpec struct {
+	// approvalDecision is the decision of the cluster admin on the Requested NonAdminBackupStorageLocation creation.
+	// +optional
+	ApprovalDecision NonAdminBSLRequest `json:"approvalDecision,omitempty"`
+}
+
+// VeleroBackupStorageLocationRequest contains information of the related Velero backup object.
```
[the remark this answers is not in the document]
mpryc: Fixed & updated corresponding oadp-operator part in that project.

internal/predicate/nonadminbackupstoragelocationrequest_predicate.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/migtools/oadp-non-admin/internal/common/function"
+)
+
+const nonAdminBackupStorageLocationRequestPredicateKey = "NonAdminBackupStorageLocationRequestPredicate"
+
+// NonAdminBackupStorageLocationRequestPredicate contains event filters for Non Admin Backup Storage Location Request objects
+type NonAdminBackupStorageLocationRequestPredicate struct{}
+
+// Create event filter accepts all NonAdminBackupStorageLocationRequest create events
+func (NonAdminBackupStorageLocationRequestPredicate) Create(ctx context.Context, evt event.CreateEvent) bool {
```
mateusoliveira43, line 33: can be deleted? only Update and Delete are used, right?

internal/handler/nonadminbackupstoragelocationrequest_handler.go
```diff
@@ -0,0 +1,73 @@
...
+
+// NonAdminBackupStorageLocationRequestHandler contains event handlers for NonAdminBackupStorageLocationRequest objects
+type NonAdminBackupStorageLocationRequestHandler struct{}
+
+// Create event handler
+func (NonAdminBackupStorageLocationRequestHandler) Create(_ context.Context, _ event.CreateEvent, _ workqueue.RateLimitingInterface) {
+	// Create event handler for the NonAdminBackupStorageLocationRequest object
+}
+
+// Update event handler adds Velero Backup's NonAdminBackup to controller queue
```
mateusoliveira43, line 40: typo

internal/predicate/nonadminbackupstoragelocationrequest_predicate.go
mpryc: The delete causes the controller to re-create, so it is handled. This is to ensure we always have corresponding Request, including situations where controller is started later.

If admin Deletes the `NonAdminBackupStorageLocationRequest` then it has lost information about previous approval and it reverts back to `pending`, deleting corresponding BSL objects.

internal/handler/nonadminbackupstoragelocationrequest_handler.go
mpryc: fixed.

internal/predicate/nonadminbackupstoragelocationrequest_predicate.go
```diff
@@ -0,0 +1,59 @@
...
+	if evt.ObjectNew.GetGeneration() != evt.ObjectOld.GetGeneration() {
+		logger.V(1).Info("Accepted Update event")
+		return true
+	}
+
+	logger.V(1).Info("Rejected Update event")
+	return false
+}
+
+// Delete event filter accepts all NonAdminBackupStorageLocation delete events
```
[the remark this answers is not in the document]
mpryc: fixed.

internal/predicate/nonadminbackupstoragelocationrequest_predicate.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Create event filter accepts all NonAdminBackupStorageLocationRequest create events
+func (NonAdminBackupStorageLocationRequestPredicate) Create(ctx context.Context, evt event.CreateEvent) bool {
+	logger := function.GetLogger(ctx, evt.Object, nonAdminBackupStorageLocationRequestPredicateKey)
+	// We don't want to accept create events for NonAdminBackupStorageLocationRequest objects, only update and delete events
+	logger.V(1).Info("Create event not accepted")
+	return false
+}
+
+// Update event filter only accepts NonAdminBackupStorageLocation update events that include spec change
```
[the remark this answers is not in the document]
mpryc: fixed.

internal/common/function/function.go
```diff
@@ -661,3 +682,34 @@ func ComputePrefixForObjectStorage(namespace, customPrefix string) string {
 	}
 	return namespace
 }
+
+// HasBSLChangesForApproval checks if there are any changes between the old and new BackupStorageLocationSpec
+// that would require admin approval. We do not want the admin to re-approve the BSL if it's just a matter of
+// the BSL creds being updated. In case of nil specs, we return true to trigger approval.
+func HasBSLChangesForApproval(oldSpec, newSpec *velerov1.BackupStorageLocationSpec) bool {
```
mateusoliveira43, line 689: is this used?

mpryc: Nice catch. It was firstly implemented when we had the update of NaBSL operation included in the design, however after talking to @shubham-pampattiwar and @weshayutin the idea changed to make things simpler and not allow update of the NaBSL and it's re-approval workflow.

Instead as in the updated design doc, the re-approval requires user to firstly delete NaBSL and then re-create new NaBSL with updated spec. It should be not a very often task to update NaBSLs by the users.

I've removed that function and relevant tests.

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -377,6 +413,180 @@ func (r *NonAdminBackupStorageLocationReconciler) setFinalizerOnNaBSL(ctx contex
...
+		updatedRejectedCondition = meta.SetStatusCondition(&nabsl.Status.Conditions, metav1.Condition{
+			Type:    string(nacv1alpha1.NonAdminBSLConditionSpecUpdateApproved),
+			Status:  metav1.ConditionFalse,
+			Reason:  "BslSpecUpdateRejected",
+			Message: message,
+		})
+		statusCondition = metav1.ConditionTrue
+		expectedPhase = nabsl.Status.Phase
+		terminalErr = reconcile.TerminalError(errors.New(message))
+	} else if nabslRequest.Status.VeleroBackupStorageLocationRequest.NACUUID == constant.EmptyString || nabslRequest.Status.VeleroBackupStorageLocationRequest.NACUUID != nabsl.Status.VeleroBackupStorageLocation.NACUUID {
```
shubham-pampattiwar, line 450: ++

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -377,6 +413,180 @@ func (r *NonAdminBackupStorageLocationReconciler) setFinalizerOnNaBSL(ctx contex
...
+			Status:  metav1.ConditionFalse,
+			Reason:  "BslSpecUpdateRejected",
+			Message: message,
+		})
+		terminalErr = reconcile.TerminalError(errors.New(message))
+		expectedPhase = nacv1alpha1.NonAdminPhaseBackingOff
+	} else {
+		switch {
+		case nabslRequest == nil:
+			shouldRequeue = true
```
shubham-pampattiwar, line 463: I think you can remove this case, right ?
This condition will always be false (GetNabslRequestByLabel will return an err in nil case)

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -377,6 +413,180 @@ func (r *NonAdminBackupStorageLocationReconciler) setFinalizerOnNaBSL(ctx contex
...
+			Type:    string(nacv1alpha1.NonAdminBSLConditionApproved),
+			Status:  statusCondition,
+			Reason:  reason,
+			Message: message,
+		})
+	}
+
+	updatePhase := updateNonAdminPhase(&nabsl.Status.Phase, expectedPhase)
+
+	if statusCondition == metav1.ConditionFalse {
```
shubham-pampattiwar, line 490: Should we also check the type of condition here ?

shubham-pampattiwar (commented):
Added a few comments, overall the flow looks good ! Thanks @mpryc !

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
```diff
@@ -0,0 +1,105 @@
...
+// +kubebuilder:validation:Enum=approve;reject;pending
+type NonAdminBSLRequest string
+
+// NonAdminBSLRequestPhase is the phase of the NonAdminBackupStorageLocationRequest
+// +kubebuilder:validation:Enum=Pending;Approved;Rejected
+type NonAdminBSLRequestPhase string
+
+// Predefined NonAdminBSLRequestConditions
+const (
+	NonAdminBSLRequestApproved NonAdminBSLRequest = "approve"
```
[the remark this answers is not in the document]
mpryc: I structured the action conditions as actionable commands to reflect the action:

 - (I) approve
 - (I) reject
 - pending <-- action to be taken

The corresponding phase names are in the past tense to reflect state (after reconcile approve spec is updated to phase status "Approved":

- Approved
- Rejected
- Pending

I can change it or we can keep the above rationale ?

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Predefined NonAdminBSLRequestPhases
+const (
+	NonAdminBSLRequestPhasePending  NonAdminBSLRequestPhase = "Pending"
+	NonAdminBSLRequestPhaseApproved NonAdminBSLRequestPhase = "Approved"
+	NonAdminBSLRequestPhaseRejected NonAdminBSLRequestPhase = "Rejected"
+)
+
+// NonAdminBackupStorageLocationRequestSpec defines the desired state of NonAdminBackupStorageLocationRequest
+type NonAdminBackupStorageLocationRequestSpec struct {
```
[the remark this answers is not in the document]
mpryc: It's in the Status, so admin can not simply generate the approval prior to it's existence - this is to reduce risk of improper labels. This spec isn't meant to be created from scratch.

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// namespace references the Namespace in which NonAdminBackupStorageLocation exists.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NonAdminBackupStorageLocationRequestStatus defines the observed state of NonAdminBackupStorageLocationRequest
+type NonAdminBackupStorageLocationRequestStatus struct {
+	// +optional
+	VeleroBackupStorageLocationRequest *VeleroBackupStorageLocationRequest `json:"veleroBackupStorageLocationRequest,omitempty"`
```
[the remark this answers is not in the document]
mpryc: About Name/Namespace - comment above, we have it here: https://github.com/migtools/oadp-non-admin/pull/216/files/30d565184e6f17c9f1a64af684ed86f2eed90018#diff-848168a85769ada3f4d261587d865e7391289a884a6a390cce8c6c6c0650740aR63-R70

This structure looks like:
 1. See nacuuid, name and namespace - it's there and it's inline with other statuses NAB or NAR - same name/namespace/nacuuid string names.
 2. The `veleroBackupStorageLocationRequest` could be just `nonAdminBslRequest` to make it shorter or remove it all together and put all of the objects  ?

```yaml
spec:
  approvalDecision: pending
status:
  phase: Pending
  veleroBackupStorageLocationRequest:
    nacuuid: non-admin-bsl-test-example-c7a90b05-9a1b-4e4c-a1e8-c3ab7f2fd4de
    name: example
    namespace: non-admin-bsl-test
    requestedSpec:
      config:
        region: eu-central-1
      credential:
        key: mycred
        name: cloud-credentials
      objectStorage:
        bucket: mpryc-nac-second
        prefix: default-nac-bucket
      provider: aws
```

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -377,6 +413,180 @@ func (r *NonAdminBackupStorageLocationReconciler) setFinalizerOnNaBSL(ctx contex
...
+		logger.V(1).Info("NonAdminBackupStorageLocationRequest already exists")
+		if updatePhaseIfNeeded(&nabslRequest.Status.Phase, nabslRequest.Spec.ApprovalDecision) {
+			if updateErr := r.Status().Update(ctx, nabslRequest); updateErr != nil {
+				logger.Error(updateErr, failedUpdateStatusError)
+				return false, updateErr
+			}
+		}
+
+		if !r.RequireApprovalForBSL && nabslRequest.Spec.ApprovalDecision != nacv1alpha1.NonAdminBSLRequestApproved {
+			logger.V(1).Info("NonAdminBackupStorageLocationRequest already exists and is not approved, updating to approved")
```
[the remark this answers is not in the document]
mpryc: yes

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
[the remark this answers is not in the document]
shubham-pampattiwar: I see, no problem, I agree with current changes.

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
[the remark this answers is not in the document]
shubham-pampattiwar: yeah I would use `nonAdminBSLRequest` because you have data in struct which pertains to NABSL and not VBSL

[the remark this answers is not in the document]
mpryc: Actually NaBSLRequest is very confusing once I made this change.

It's because the NaBSLRequest is the object itself (Request object) and here in the status we are refering to nonAdminBsl object spec which was origin for our NaBSLRequest.

Maybe we call it full name, so it will look like

```yaml
spec:
  approvalDecision: pending
status:
  phase: Pending
  nonAdminBackupStorageLocation:
    nacuuid: non-admin-bsl-test-example-c7a90b05-9a1b-4e4c-a1e8-c3ab7f2fd4de
    name: example
    namespace: non-admin-bsl-test
    requestedSpec:
      config:
        region: eu-central-1
      credential:
        key: mycred
        name: cloud-credentials
      objectStorage:
        bucket: mpryc-nac-second
        prefix: default-nac-bucket
      provider: aws
```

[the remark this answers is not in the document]
mpryc: @shubham-pampattiwar with the latest modification where I chosen `nonAdminBackupStorageLocation` the NonAdminBackupStorageLocationRequest object is rendered as follows, let me know if we want to change it or keep it:

```yaml
[...]
spec:
  approvalDecision: pending
status:
  nonAdminBackupStorageLocation:
    nacuuid: non-admin-bsl-test-nonadmi-09d36eb9-381f-4cad-bef6-1a354af58535
    name: nonadminbackupstoragelocation-sample
    namespace: non-admin-bsl-test
    requestedSpec:
      config:
        region: eu-central-1
      credential:
        key: mycred
        name: cloud-credentials
      objectStorage:
        bucket: mpryc-nac-second
        prefix: default-nac-bucket
      provider: aws
  phase: Pending
```

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: Correct, actually `GetNabslRequestByLabel` is not returning error if nothing is found, works same in other Backup/Restore/NAB objects so don't want to change how the GetNabslRequestByLabel works, however this check is not required and added new one on the nabslRequest nil just afte GetNabslRequestByLabel so other parts of the function won't have nil object (which will in theory never happen, but better safe then sorry)

internal/controller/nonadminbackupstoragelocation_controller.go
mpryc: No need as we have full control over this variable and we set this variable on line:
https://github.com/migtools/oadp-non-admin/pull/216/files/30d565184e6f17c9f1a64af684ed86f2eed90018#diff-ab1df5cab3611d202a4b22ab31a471f15bd9b010c803a8df03b55ce6b7e09c2eR433

So it's not value which we get from the cluster from the condition, but we know for which condition we are setting it to true/false.

I've renamed this var to be more informative: `adminApprovedCondition`

internal/controller/nonadminbackupstoragelocation_controller.go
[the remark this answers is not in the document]
mpryc: Updated

mpryc:
@shubham-pampattiwar I think I addressed all of your comments and updated oadp-operator, please check them. Some of the comments I've implemented slightly different then you asked and gave rationale behind my implementation changes.

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -377,6 +413,180 @@ func (r *NonAdminBackupStorageLocationReconciler) setFinalizerOnNaBSL(ctx contex
...
+
+	if !reflect.DeepEqual(nabslRequest.Status.NonAdminBackupStorageLocationRequestStatusInfo.DeepCopy().RequestedSpec, nabsl.Spec.BackupStorageLocationSpec) {
+		message = "NaBSL Spec update not allowed. Changes will not be applied. Delete NaBSL and create new one with updated spec"
+		updatedRejectedCondition = meta.SetStatusCondition(&nabsl.Status.Conditions, metav1.Condition{
+			Type:    string(nacv1alpha1.NonAdminBSLConditionSpecUpdateApproved),
+			Status:  metav1.ConditionFalse,
+			Reason:  "BslSpecUpdateRejected",
+			Message: message,
+		})
+		adminApprovedCondition = metav1.ConditionTrue
```
mateusoliveira43, line 450: since this error case, shouldn't it be false and BackingOff?

mpryc: No.

In this scenario we have two conditions:

```go
	NonAdminBSLConditionApproved           NonAdminBSLCondition = "ClusterAdminApproved"
	NonAdminBSLConditionSpecUpdateApproved NonAdminBSLCondition = "SpecUpdateApproved"
```

This flag however is not used for updating any of the above conditions from (line number 480), it is set to true to not delete previously created BSL nor Secret on line number 490. This was added here and I can introduce new flag e.g. `deleteNacBslSecret`, but thought to use it to not have too many vars in this logic

mpryc: added new flag `preserveVeleroBslSecret` in the go code, to make it easier to read the workflow. Thanks.

mpryc: https://github.com/migtools/oadp-non-admin/pull/216/commits/3cf8e01c6a3d93bc6c0376fc390d136b6d2132af

mpryc:
>   would remove it or change message to something like `automatically approved`

We would add the extra logic to distinguish between different approval types (human & non-human) which would add complexity in some cases when one is switched to another. Plus it would expose DPA setting to the users with this message, so I think either we think of better generic wording or stick with what's there.

> * if admin changes from no approval to approval, nothing happens. Should all NABSL be set to pending approval? (The other way around, when admin changes from approval to no approval, seems to work, i.e., all NABSL go to approved state)

This is actually desired. If the admin want's to re-approve all of the NaBSLs after toggling this to true, then the easiest is to remove all of the NonAdminBackupStorageLocationRequest objects from the openshift-adp namespace:
[ - If there was no approval required, switching it to on will NOT delete all of the previously approved 
](https://github.com/migtools/oadp-non-admin/blob/3cf8e01c6a3d93bc6c0376fc390d136b6d2132af/docs/design/non_admin_backupstoragelocation.md#enabling-bsl-approval-request-feature)

Vice-versa:
https://github.com/migtools/oadp-non-admin/blob/3cf8e01c6a3d93bc6c0376fc390d136b6d2132af/docs/design/non_admin_backupstoragelocation.md#disabling-bsl-approval-request-feature

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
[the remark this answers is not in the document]
shubham-pampattiwar: `nonAdminBackupStorageLocation` sounds good !

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// namespace references the Namespace in which NonAdminBackupStorageLocation exists.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NonAdminBackupStorageLocationRequestStatus defines the observed state of NonAdminBackupStorageLocationRequest
+type NonAdminBackupStorageLocationRequestStatus struct {
+	// +optional
+	NonAdminBackupStorageLocationRequestStatusInfo *NonAdminBackupStorageLocationRequestStatusInfo `json:"nonAdminBackupStorageLocation,omitempty"`
```
shubham-pampattiwar, line 75: I though we wanted this as `nonAdminBackupStorageLocation` ?

internal/controller/nonadminbackupstoragelocation_controller.go
```diff
@@ -377,6 +413,182 @@ func (r *NonAdminBackupStorageLocationReconciler) setFinalizerOnNaBSL(ctx contex
...
+		logger.V(1).Info("NonAdminBackupStorageLocationRequest already exists")
+		if updatePhaseIfNeeded(&nabslRequest.Status.Phase, nabslRequest.Spec.ApprovalDecision) {
+			if updateErr := r.Status().Update(ctx, nabslRequest); updateErr != nil {
+				logger.Error(updateErr, failedUpdateStatusError)
+				return false, updateErr
+			}
+		}
+
+		if !r.RequireApprovalForBSL && nabslRequest.Spec.ApprovalDecision != nacv1alpha1.NonAdminBSLRequestApproved {
+			logger.V(1).Info("Unapproved NonAdminBackupStorageLocationRequest found; approving as requireApprovalForBSL on the DPA is not true.")
```
shubham-pampattiwar, line 543: ++

api/v1alpha1/nonadminbackupstoragelocationrequest_types.go
mpryc: Including structure of `NonAdminBackupStorageLocation` will look wird imo, because it contains yet another status which contains Velero BSL:
https://github.com/migtools/oadp-non-admin/blob/master/api/v1alpha1/nonadminbackupstoragelocation_types.go#L40

mpryc: That's just go var name, which isn't used in the api/v1alpha1, right? If I used the `NonAdminBackupStorageLocation` then we had this object already https://github.com/migtools/oadp-non-admin/blob/121987f57ecb24d4bc65e4a2bef81593525a5a64/api/v1alpha1/nonadminbackupstoragelocation_types.go#L74

shubham-pampattiwar (approved):

shubham-pampattiwar (approved):

mateusoliveira43 (approved):

weshayutin:
VERY WELL DONE FOLKS.. this is working for me as expected.
I have not checked for dupes yet.. but created two issues

https://github.com/migtools/oadp-non-admin/issues/232
https://github.com/migtools/oadp-non-admin/issues/231

Issue #202, opened 2025-02-18
Restrict non-default VSL specification for NAB spec


shubham-pampattiwar:

mateusoliveira43:
Can VSL be set as default? I do not see this option in VSL spec

shubham-pampattiwar:
yeah we dont have a VSL spec for it. I think what we want to do here is if user is not using a default BSL i.e. they have NABSL configured, then in such a scenario we should restrict the usage of VSL spec altogether.
Does that make sense ?

mateusoliveira43:
I think is more than that

I think we can not enforce in admin side the value of VSL in backup spec

I think we can not allow non admin to add this value in NAB spec (velero autocompletes in backup side)

In this last case, I do not think is possible to tell velero to not autocomplete the value if use is using NABSL

PR #218, opened 2025-02-24
fix: restrict VSL spec in NAB

mateusoliveira43:
## Why the changes were made

Fix https://github.com/migtools/oadp-non-admin/issues/202

Check that a NAB can not have volumeSnapshotLocations spec set.

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -135,6 +135,10 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 		}
 	}
 
+	if nonAdminBackup.Spec.BackupSpec.VolumeSnapshotLocations != nil {
+		return fmt.Errorf(constant.NABRestrictedErr, "spec.backupSpec.volumeSnapshotLocations")
+	}
+
 	enforcedSpec := reflect.ValueOf(enforcedBackupSpec).Elem()
 	for index := range enforcedSpec.NumField() {
 		enforcedField := enforcedSpec.Field(index)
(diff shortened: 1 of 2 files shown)


sseago (approved):

internal/common/function/function.go
```diff
@@ -135,6 +135,10 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 		}
 	}
 
+	if nonAdminBackup.Spec.BackupSpec.VolumeSnapshotLocations != nil {
```
mpryc, line 138: Should there be also a check if `&& len(nonAdminBackup.Spec.BackupSpec.VolumeSnapshotLocations) > 0` ?

mateusoliveira43: I think current implementation does not allow non admin user to set `spec.backupSpec.volumeSnapshotLocations`

with your idea, it would allow to set to an empty list (but velero coudl still propably change the value on backup side)

which is better?

mpryc: Ok your is better as it needs to be nil.

One last bit - will this be still nil if the sync controller recreates it from s3 Backup object spec?

mateusoliveira43: no, only special case is BSL ref that is updated https://github.com/migtools/oadp-non-admin/blob/2d5e4e7ae5ce34063aa3d9e49c399cc8a8818870/internal/controller/nonadminbackupsynchronizer_controller.go#L148-L153

mpryc (approved):

Issue #192, opened 2025-02-14
We should also validate that Velero BSL backupSyncPeriod is not greater or equal to nonAdmin.backupSyncPeriod


shubham-pampattiwar:
TODO mentioned https://github.com/openshift/oadp-operator/pull/1624

mateusoliveira43:
Scenarios

1. default BSL in OADP namespace
    This is the hard part. If default BSL was created from DPA, we could add validation here https://github.com/openshift/oadp-operator/blob/580a0524836f4a215504ab57aca028f01f87a45a/internal/controller/bsl.go#L21
    But if default BSL was created standalone, we would need to write validation on Velero side

2. NABSL in non admin namespace
    This is the easy fix, we already get the value admin configured here (or the default was applied) https://github.com/migtools/oadp-non-admin/blob/3c69c1c137ab60897f6e73c2a34035b370cea874/cmd/main.go#L207 
    We would need to just add validation here https://github.com/migtools/oadp-non-admin/blob/3c69c1c137ab60897f6e73c2a34035b370cea874/internal/common/function/function.go#L200 to check if resulting BSL sync period is not greater or equal to DPA value

(we talked today around NABSL/BSL scenarios, do you think more scenarios will appear before this first release?)

How should this issue be fixed:

1. applying validation on NABSLs and default BSL created from DPA
2. applying validation on NABSLs

mateusoliveira43:
possibly related https://github.com/migtools/oadp-non-admin/issues/204

PR #223, opened 2025-02-24
fix: NABSL sync period validation

mateusoliveira43:
## Why the changes were made

https://github.com/migtools/oadp-non-admin/issues/192

Create NABSL with backup sync period greater or equal to NAC backup sync period. NABSL should error out

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"time"
 
 	// TODO when to update oadp-operator version in go.mod?
 	"github.com/openshift/oadp-operator/api/v1alpha1"
@@ -139,7 +140,7 @@ func main() {
 
 	restConfig := ctrl.GetConfigOrDie()
 
-	dpaConfiguration, err := getDPAConfiguration(restConfig, oadpNamespace)
+	dpaConfiguration, defaultSyncPeriod, err := getDPAConfiguration(restConfig, oadpNamespace)
 	if err != nil {
 		setupLog.Error(err, "unable to get enforced spec")
 		os.Exit(1)
@@ -197,6 +198,8 @@ func main() {
 		Scheme:                mgr.GetScheme(),
 		OADPNamespace:         oadpNamespace,
 		RequireApprovalForBSL: *dpaConfiguration.RequireApprovalForBSL,
+		SyncPeriod:            dpaConfiguration.BackupSyncPeriod.Duration,
+		DefaultSyncPeriod:     defaultSyncPeriod,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to setup NonAdminBackupStorageLocation controller with manager")
 		os.Exit(1)
@@ -242,7 +245,7 @@ func main() {
 	}
 }
 
-func getDPAConfiguration(restConfig *rest.Config, oadpNamespace string) (v1alpha1.NonAdmin, error) {
+func getDPAConfiguration(restConfig *rest.Config, oadpNamespace string) (v1alpha1.NonAdmin, *time.Duration, error) {
 	dpaConfiguration := v1alpha1.NonAdmin{
 		GarbageCollectionPeriod: &metav1.Duration{
 			Duration: v1alpha1.DefaultGarbageCollectionPeriod,
@@ -254,6 +257,7 @@ func getDPAConfiguration(restConfig *rest.Config, oadpNamespace string) (v1alpha
 		EnforceRestoreSpec:    &velerov1.RestoreSpec{},
 		RequireApprovalForBSL: ptr.To(false),
 	}
+	var defaultSyncPeriod *time.Duration
 
 	dpaClientScheme := runtime.NewScheme()
 	utilruntime.Must(v1alpha1.AddToScheme(dpaClientScheme))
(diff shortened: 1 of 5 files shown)


internal/common/function/function.go
```diff
@@ -219,6 +220,25 @@ func ValidateBslSpec(ctx context.Context, clientInstance client.Client, nonAdmin
...
+	bslSyncPeriodErrorMessage := "NABSL spec.backupStorageLocationSpec.backupSyncPeriod (%v) can not be greater or equal non admin backupSyncPeriod (%v)"
+	if nonAdminBsl.Spec.BackupStorageLocationSpec.BackupSyncPeriod != nil {
+		if appliedBackupSyncPeriod <= nonAdminBsl.Spec.BackupStorageLocationSpec.BackupSyncPeriod.Duration {
+			return fmt.Errorf(
+				bslSyncPeriodErrorMessage,
+				nonAdminBsl.Spec.BackupStorageLocationSpec.BackupSyncPeriod.Duration, appliedBackupSyncPeriod,
+			)
+		}
+	} else {
+		// TODO how to get user configured default backup sync period?
```
mateusoliveira43, line 232: this can give false positives

if admin set defaultBackupSyncPeriod though DPA velero args, we do not know what is the value being used

shubham-pampattiwar: check for velero container args and push those values to NAC deployment and then use them. How does that sound ?

mateusoliveira43: added velero args to function

shubham-pampattiwar (commented):
needs rebase

cmd/main.go
```diff
@@ -254,9 +258,12 @@ func getDPAConfiguration(restConfig *rest.Config, oadpNamespace string) (v1alpha
 			if nonAdmin.BackupSyncPeriod != nil {
 				dpaConfiguration.BackupSyncPeriod.Duration = nonAdmin.BackupSyncPeriod.Duration
 			}
+			if dpa.Spec.Configuration.Velero.Args != nil && dpa.Spec.Configuration.Velero.Args.BackupSyncPeriod != nil {
+				defaultSyncPeriod = dpa.Spec.Configuration.Velero.Args.BackupSyncPeriod
```
mpryc, line 262: In case this is not met, we may have nil point in the caller of this function, because we will have dpaConfiguration, but defaultSyncPeriod was never initialized from line 231.

mateusoliveira43: I think this is ok

I want to it to be nil to check it here https://github.com/migtools/oadp-non-admin/pull/223/files#diff-de455907ed1c5413fa6248ebea483af559585f8f6277a7c8b370813b9bb254f9R232

or does this break usage?

internal/common/function/function.go
```diff
@@ -219,6 +220,32 @@ func ValidateBslSpec(ctx context.Context, clientInstance client.Client, nonAdmin
...
+				bslSyncPeriodErrorMessage,
+				nonAdminBsl.Spec.BackupStorageLocationSpec.BackupSyncPeriod.Duration, appliedBackupSyncPeriod,
+			)
+		}
+	} else {
+		if defaultBackupSyncPeriod != nil {
+			if appliedBackupSyncPeriod <= *defaultBackupSyncPeriod {
+				return fmt.Errorf(
+					bslSyncPeriodErrorMessage,
+					defaultBackupSyncPeriod, appliedBackupSyncPeriod,
```
mpryc, line 236: Is this right? The `defaultBackupSyncPeriod` is a pointer and the `appliedBackupSyncPeriod` is not, so should this be used as pointer `*defaultBackupSyncPeriod` ?

cmd/main.go
mpryc: I think it may break us if the DPA is set without default BSL (so no BSL at all in the cluster and DPA reconciled), then we won't ValidateBSL properly, need to test this scenario.

internal/common/function/function.go
mateusoliveira43: I think is ok, Go renders message correctly
```
NABSL spec.backupStorageLocationSpec.backupSyncPeriod (1h40m0s) can not be greater or equal non admin backupSyncPeriod (30s)
```

mpryc (changes requested):
I am hitting the following error:
```shell
NABSL spec.backupStorageLocationSpec.backupSyncPeriod (1m0s) can not be greater or equal non admin backupSyncPeriod (1m0s)
```

Steps to reproduce:
 1. Create DPA with the following spec:
```yaml
spec:
  backupImages: false
  configuration:
    nodeAgent:
      enable: true
      uploaderType: kopia
    velero:
      defaultPlugins:
        - openshift
        - aws
      noDefaultBackupLocation: true
      resourceTimeout: 10m
  nonAdmin:
    backupSyncPeriod: 1m
    enable: true
    requireApprovalForBSL: true
```

2. Create NonAdminBSL with the following spec (no backupSyncPeriod specified):
```yaml
spec:
  backupStorageLocationSpec:
    config:
      region: eu-central-1
    credential:
      key: mycred
      name: cloud-credentials
    objectStorage:
      bucket: mpryc-nac-second
      prefix: default-nac-bucket
    provider: aws
```

Expected behavior would be to allow BSL creation, because user did not specify one.
Current behavior: Error on the NaBSL object.

mpryc (approved):

mpryc:
The default velero is 1minute so we are fine here as the DPA admin can set it equal meaning each user will have to specify other then default to have greater value.

PR #243, opened 2025-02-27
vscode sample settings.json

kaovilai:
Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

### DIFF
--- a/.vscode.sample/settings.json
+++ b/.vscode.sample/settings.json
@@ -1,0 +1,6 @@
+{
+    "go.lintTool": "golangci-lint",
+    "go.alternateTools": {
+        "golangci-lint": "${workspaceFolder}/bin/golangci-lint-v1.62.2" // corresponds to golangci-lint in Makefile
+    },
+}


.vscode.sample/settings.json
mateusoliveira43, line 1: this is IDE specific

I would not add this to the repo

whoever wants to work on this project does not need to use VSCode

kaovilai: what about https://github.com/migtools/oadp-non-admin/blob/47e640d18dde2505dbfaa9d143095167d308f3b3/.editorconfig#L1-L20

kaovilai: otherwise my IDE is throwing non matching linter errors due to newer golangci-lint picked up by IDE. I would still like to use IDE automatic linting integration with golangci-lint

kaovilai: How about https://github.com/openshift/oadp-operator/blob/a6262b410c88be06226afebf1c1caf52bf0d20d2/docs/developer/testing/AttatchingIDEDebuggers.md?plain=1#L5 ?

mpryc: I don't think it's unreasonable to add `.vscode/settings.json` file to the project repository with some sensible defaults and at the same time ensure .gitignore contains this, so developers won't be adding their generated by the IDE changes. What I am wondering if it's possible to add it in a way it won't be pinning to a particular version (e.g. via the hack/<somelintscript>) and use it in all the places, so we don't have to remember to bump the golangci binary every time. Here are few examples that already are within OpenShift org:

https://github.com/openshift/api/blob/635291d6fdf1c4f4cb2b94b460f8d4f9cc377571/.vscode/settings.json#L4

https://github.com/openshift/console/blob/e9d6ead6852600993cffef3d50cb9b122d64c068/frontend/.vscode/settings.json#L4

https://github.com/openshift/cluster-network-operator/blob/d09028dfda8c632e4c4730319b9f07f0ac92d82a/.vscode/settings.json#L3

PR #249, opened 2025-03-04
Check the NaBSL status and conditions before creating Velero Backup

### LINKED ISSUE #246: Able to perform Backup successsfully using NABSL in phase: BackingOff
## Description

I created the DPA spec, without using any enforceBSLSpec and created a NABSL with provider aws.
Then I edited the DPA, and now I am enforcing provider: gcp using enforceBSLSpec from the DPA spec.
So the Old NABSL now has this condition
```
"the administrator has restricted spec.backupStorageLocationSpec.provider field to: gcp'"
```
and its Phase is BackingOff, but I can still very well perform Backup Successfully using the NABSL.

## How to reproduce

1. Create DPA without any enforceBSLSpec
2. Create a NABSL in non admin user namespace, with provider lets say aws
3. Now, add enforceBSLSpec by editing DPA and set provider: gcp 
4. Now Check NABSL has the error condition and phase:BackingOff
5. Try to perform backup using the same NABSL.

DPA spec after enforcement:

```yaml
spec:
    backupLocations:
    - velero:
        credential:
          key: cloud
          name: cloud-credentials-aws
        default: true
        objectStorage:
          bucket: oadp112571t7jvn
          prefix: velero
        provider: aws
    configuration:
      nodeAgent:
        enable: true
        uploaderType: kopia
      velero:
        defaultPlugins:
        - csi
        - aws
        - openshift
    nonAdmin:
      enable: true
      enforceBSLSpec:
        provider: gcp
```

NABSL after putting enforcement:

```yaml
metadata:
  creationTimestamp: "2025-03-01T08:36:19Z"
  finalizers:
  - nonadminbackupstoragelocation.oadp.openshift.io/finalizer
  generation: 1
  name: nabsl1
  namespace: test-nac-1
  resourceVersion: "81708"
  uid: f8fb1f1b-2258-47ea-9b99-14c3f6c32c71
spec:
  backupStorageLocationSpec:
    credential:
      key: cloud
      name: cloud-credentials-aws
    objectStorage:
      bucket: oadp112571t7jvn
      prefix: velero
    provider: aws
status:
  conditions:
  - lastTransitionTime: "2025-03-01T09:16:54Z"
    message: 'the administrator has restricted spec.backupStorageLocationSpec.provider
      field to: gcp'
    reason: BslSpecValidation
    status: "False"
    type: Accepted
  - lastTransitionTime: "2025-03-01T08:36:19Z"
    message: NonAdminBackupStorageLocationRequest approval decision set to Approve
    reason: BslSpecApproved
    status: "True"
    type: ClusterAdminApproved
  - lastTransitionTime: "2025-03-01T08:36:19Z"
    message: Secret successfully created in the OADP namespace
    reason: SecretCreated
    status: "True"
    type: SecretSynced
  - lastTransitionTime: "2025-03-01T08:36:19Z"
    message: BackupStorageLocation successfully created in the OADP namespace
    reason: BackupStorageLocationCreated
    status: "True"
    type: BackupStorageLocationSynced
  phase: BackingOff
  veleroBackupStorageLocation:
    nacuuid: test-nac-1-nabsl1-c43a32d4-db3a-40e6-98cb-cb3ea132a13b
    name: test-nac-1-nabsl1-c43a32d4-db3a-40e6-98cb-cb3ea132a13b
    namespace: openshift-adp
    status:
      lastSyncedTime: "2025-03-01T09:15:54Z"
      lastValidationTime: "2025-03-01T09:15:54Z"
      phase: Available
```

Backup can be performed successfully as BSL remains available.

mpryc:
Fixes #246

Ensure the state of the NaBSL is satisfying the Backup operation, in a case the admin redefined enforcement the NaBSL state was correctly synced back from the enforcement, however we disallowed real updated of the Velero BSL object causing NaBSL being in not allowed state, however previous Velero BSL fully functional.

This logic fixes that issue.

## Why the changes were made

To fix #246 

1. As described in the #246 issue:
2. Create enforcements for the NaBSL in the DPA
3. Create NaBSL to satisfy enforcements
4. Change the NaBSL enforcements in the DPA
5. Try to perform Backup operation using the NaBSL

Expected result: Backup can not be created.

### DIFF
--- a/internal/common/function/function.go
+++ b/internal/common/function/function.go
@@ -121,6 +121,15 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 		} else if err != nil {
 			return fmt.Errorf("NonAdminBackup spec.backupSpec.storageLocation is invalid: %v", err)
 		}
+		if nonAdminBsl.Status.Phase != nacv1alpha1.NonAdminPhaseCreated {
+			return fmt.Errorf("NonAdminBackupStorageLocation is not in created state and can not be used for the NonAdminBackup")
+		}
+
+		for _, condition := range nonAdminBsl.Status.Conditions {
+			if condition.Status != metav1.ConditionTrue {
+				return fmt.Errorf("NonAdminBackupStorageLocation has an unsatisfied condition: %s is %s", condition.Type, condition.Status)
+			}
+		}
 		if nonAdminBsl.Status.VeleroBackupStorageLocation == nil || nonAdminBsl.Status.VeleroBackupStorageLocation.NACUUID == constant.EmptyString {
 			return fmt.Errorf("unable to get VeleroBackupStorageLocation UUID from NonAdminBackupStorageLocation Status")
 		}
--- a/internal/common/function/function_test.go
+++ b/internal/common/function/function_test.go
@@ -407,6 +407,7 @@ func TestValidateBackupSpecEnforcedFields(t *testing.T) {
 						Namespace: "self-service-namespace",
 					},
 					Status: nacv1alpha1.NonAdminBackupStorageLocationStatus{
+						Phase: nacv1alpha1.NonAdminPhaseCreated,
 						VeleroBackupStorageLocation: &nacv1alpha1.VeleroBackupStorageLocation{
 							NACUUID: "user-defined-backup-storage-location-uuid",
 						},
--- a/internal/controller/nonadminbackup_controller_test.go
+++ b/internal/controller/nonadminbackup_controller_test.go
@@ -52,6 +52,7 @@ type nonAdminBackupSingleReconcileScenario struct {
 	resultError                         error
 	nonAdminBackupPriorStatus           *nacv1alpha1.NonAdminBackupStatus
 	nonAdminBackupSpec                  nacv1alpha1.NonAdminBackupSpec
+	nonAdminBackupStorageLocationStatus *nacv1alpha1.NonAdminBackupStorageLocationStatus
 	nonAdminBackupExpectedStatus        nacv1alpha1.NonAdminBackupStatus
 	result                              reconcile.Result
 	createVeleroBackup                  bool
@@ -336,7 +337,8 @@ var _ = ginkgo.Describe("Test single reconciles of NonAdminBackup Reconcile func
 					},
 				}
 				gomega.Expect(k8sClient.Create(ctx, nonAdminBackupStorageLocation)).To(gomega.Succeed())
-				nonAdminBackupStorageLocation.Status.Phase = nacv1alpha1.NonAdminPhaseCreated
+
+				nonAdminBackupStorageLocation.Status = *scenario.nonAdminBackupStorageLocationStatus
 				gomega.Expect(k8sClient.Status().Update(ctx, nonAdminBackupStorageLocation)).To(gomega.Succeed())
 
 				nonAdminBackup.Spec.BackupSpec.StorageLocation = nonAdminBackupStorageLocationName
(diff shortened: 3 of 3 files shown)


internal/common/function/function.go
```diff
@@ -121,6 +121,15 @@ func ValidateBackupSpec(ctx context.Context, clientInstance client.Client, oadpN
 		} else if err != nil {
 			return fmt.Errorf("NonAdminBackup spec.backupSpec.storageLocation is invalid: %v", err)
 		}
+		if nonAdminBsl.Status.Phase != nacv1alpha1.NonAdminPhaseCreated {
+			return fmt.Errorf("NonAdminBackupStorageLocation is not in created state and can not be used for the NonAdminBackup")
+		}
+
+		for _, condition := range nonAdminBsl.Status.Conditions {
```
mateusoliveira43, line 128: I do not like this approach. If a new condition is added, this can break us

If there is a certain condition that is important to check, we can check it, but checking all seems the wrong approach

shubham-pampattiwar: @mpryc the bug should be fixed via only checking the phase, right ?

kaovilai:
```
// Condition contains details for one aspect of the current state of this API Resource.
// ---
// This struct is intended for direct use as an array at the field path .status.conditions.  For example,
//
//	type FooStatus struct{
//	    // Represents the observations of a foo's current state.
//	    // Known .status.conditions.type are: "Available", "Progressing", and "Degraded"
//	    // +patchMergeKey=type
//	    // +patchStrategy=merge
//	    // +listType=map
//	    // +listMapKey=type
//	    Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
//
//	    // other fields
//	}
```

[condition](https://github.com/kubernetes/apimachinery/blob/v0.32.2/pkg/apis/meta/v1/types.go#L1589C1-L1603C5) is not always something that needs to be true to be satisfied.

We are establishing that all conditions in this repo if true is considered an unsatisfied condition. Is that what we want?

mpryc: I understand your concern about checking all conditions, but in our specific case, I believe it's a safer approach. I can rework this logic to check for 4 conditions individually:

https://github.com/migtools/oadp-non-admin/blob/af1bbcd371bfd7c7117b0aca93105e16c7ddefb1/api/v1alpha1/nonadminbackupstoragelocation_types.go#L30-L33

Each of these conditions must be satisfied to ensure that a NaBSL is actually usable. Relying solely on the status field could lead to situations where an NaBSL is marked as "OK" but is still not in a functional state.

By explicitly checking these conditions, we reduce potential issues that could arise if a new phase or status logic is introduced in the future. I believe this approach is a safeguard rather than an unnecessary restriction, but as I wrote I can change this to explicitly check for the above 4 conditions instead of iterating through all of them.

mpryc: Added only 3 conditions which in my opinion are a *must* to be met here before continuing backup and none of those 3 conditions can be false.

mateusoliveira43: I think logic is not correct

current we iterate through NABSL status conditions, if that condition is one of the 3 and is false, we error, right?

but if for example NABSL status conditions has no conditions, no problem happens, but it should report a problem, no?

would change logic to iterate through required conditions types, check if that condition is present in NABSL status (error if not) and if it true (error if not). Example implementation
```go
		requiredConditions := []string{
			string(nacv1alpha1.NonAdminBSLConditionSecretSynced),
			string(nacv1alpha1.NonAdminBSLConditionBSLSynced),
			string(nacv1alpha1.NonAdminBSLConditionApproved),
		}
		for _, condition := range requiredConditions {
			conditionInNABSL := meta.FindStatusCondition(nonAdminBsl.Status.Conditions, condition)
			if conditionInNABSL == nil {
				return fmt.Errorf("TODO")
			}
			if conditionInNABSL.Status != metav1.ConditionTrue {
				return fmt.Errorf("TODO")
			}
		}
```

mpryc: @shubham-pampattiwar updated the implementation to relay only on phase, which should be sufficient.

mateusoliveira43 (approved):

shubham-pampattiwar (approved):
/lgtm

Issue #259, opened 2025-04-01
Pin OADP Operator version in go.mod


shubham-pampattiwar:

kaovilai:
#271

PR #271, opened 2025-04-25
Update OADP Operator dep to 1.5

shubham-pampattiwar:
## Why the changes were made

Fixes https://github.com/migtools/oadp-non-admin/issues/259


go.mod
```diff
@@ -9,10 +9,10 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
-	github.com/openshift/oadp-operator v1.0.2-0.20250227152435-9267da21ab64
+	github.com/openshift/oadp-operator v1.0.2-0.20250425033753-5a4f3ed74119
```
mateusoliveira43, line 12: better to wait code freeze to update here?

mpryc: IMO it's easy to bump twice so the current deps can be tested as early as possible without code changes.

